PR #7, opened 2020-11-26
Small fixes and a lot of TODOs for DataGeneratorAllInchikeys

svenvanderburg:
Here is my review for the DataGeneratorAllInchikeys class.
It includes some changes that make it a bit easier to read (and more reproducable) and a lot of TODOs.
I think most of the code looks sane, but I am maybe 80% sure because there is some missing documentation that makes it quite hard for me to really grasp everything. I suggest that you @florian-huber add some more documentation (basically following the TODOs that I put in there), where the goal is to make it clearer what the code does. We could also do some pair coding on the TODOs together.

I suggest that we see what we want to improve for this class and then apply that to the other DataGenerator classes.


ms2deepscore/data_generators.py
```diff
@@ -138,35 +158,31 @@ def __data_augmentation(self, document_dict):
...
-
-            # Add scaled weight to right bins
-            X[1][i, query_idx] = query_values ** self.peak_scaling
-            y[i] = self.score_array[IDs[0], IDs[1]]
-            if np.isnan(y[i]):
-                y[i] = np.random.random(1)
+        for i_batch, pair in enumerate(batch_ids):
+            for i_pair, id in enumerate(pair):
+                inchikey = self.inchikey_mapping.loc[id]["inchikey"]
+                spectrum_id = np.random.choice(np.where(self.inchikeys_array == inchikey)[0])
```
svenvanderburg, line 175: I am not sure if I have the names of ids correct now. There seem to be multiple levels in the hierarchy: training_sample_id -> inchikey -> spectrum_id . Is that correct? I think it would be nice to use these (or better names throughout)

ms2deepscore/data_generators.py
```diff
@@ -4,15 +4,19 @@
 import pandas as pd
 from tensorflow.keras.utils import Sequence
 
+# Set random seed for reproducibility
+np.random.seed(42)
```
svenvanderburg, line 8: This will ensure anyone trying to reproduce your results will get exactly the same!

ms2deepscore/data_generators.py
```diff
@@ -138,35 +158,31 @@ def __data_augmentation(self, document_dict):
...
+                spectrum_id = np.random.choice(np.where(self.inchikeys_array == inchikey)[0])
+                idx, values = self.__data_augmentation(self.spectrums_binned_dicts[spectrum_id])
+                X[i_pair][i_batch, idx] = values ** self.peak_scaling
+
+            y[i_batch] = self.score_array[pair[0], pair[1]]
+
+            # TODO: Add documentation. This means if there is no label we just add a random one
+            #  right? Maybe it would be nicer to not include these (which would mean completely
+            #  rearranging everything)
+            if np.isnan(y[i_batch]):
```
svenvanderburg, line 184: Can this actually happen? Given that in line 62 (of the old file) you set nan values to 0?

ms2deepscore/data_generators.py
```diff
@@ -4,33 +4,41 @@
...
         with a randomly chosen other spectrum that corresponds to a reference score
         as defined in same_prob_bins.
 
         Parameters
         ----------
         spectrums_binned_dicts
+            TODO: Make this less nested and clearer what I should exactly pass.
             List of dictionaries with the binned peak positions and intensities.
-        list_IDs
+        training_ids
```
florian-huber, line 35: or ``compound_ids``

ms2deepscore/data_generators.py
florian-huber: was changed

ms2deepscore/data_generators.py
florian-huber: has been changed in PR #8

ms2deepscore/data_generators.py
florian-huber: Good point. thanks!

ms2deepscore/data_generators.py
florian-huber: changed it in PR #8

florian-huber (approved):
Thanks a lot @svenvanderburg for the review!

PR #8, opened 2020-11-26
Refactor data generators

florian-huber:
- refactoring of the two main generators used
- they still have a lot of code duplication, but since I am not sure which one will make it I would leave this for later
- added some first tests (currently mostly only checking that the generators return something of right shape)


tests/test_data_generators.py
```diff
@@ -0,0 +1,122 @@
+import os
+import json
+import numpy as np
+import pytest
+from matchms import Spectrum
+from ms2deepscore.data_generators import DataGeneratorAllInchikeys
+from ms2deepscore.data_generators import DataGeneratorAllSpectrums
+path_tests  = os.path.dirname(__file__)
```
svenvanderburg, line 8: Should be a constant.

tests/test_data_generators.py
```diff
@@ -0,0 +1,122 @@
...
+                                    inchikey_score_mapping, inchikeys_all,
+                                    dim=dimension, batch_size=batch_size,
+                                    num_turns=num_turns, peak_scaling=peak_scaling,
+                                    shuffle=True, ignore_equal_pairs=True,
+                                    same_prob_bins=same_prob_bins,
+                                    augment_peak_removal_max=0.0,
+                                    augment_peak_removal_intensity=0.0,
+                                    augment_intensity=0.0)
+    assert "inchikeys_all and spectrums_binned must have the same dimension." in str(msg.value), \
+        "Expected different expection to be raised"
```
svenvanderburg, line 122:
```suggestion
        "Expected different exception to be raised"
```

ms2deepscore/data_generators.py
```diff
@@ -10,150 +11,188 @@
...
-        listed in *training_ids* num_turns times in every epoch. It will then randomly
-        pick one the spectra corresponding to this InchiKey (if multiple) and pair it
-        with a randomly chosen other spectrum that corresponds to a reference score
-        as defined in same_prob_bins.
 
         Parameters
         ----------
-        spectrums_binned_dicts
+        spectrums_binned
             TODO: Make this less nested and clearer what I should exactly pass.
```
svenvanderburg, line 36: I guess this fixes the TODO?

ms2deepscore/data_generators.py
```diff
@@ -10,150 +11,188 @@
...
+        self.augment_peak_removal_max = augment_peak_removal_max
+        self.augment_peak_removal_intensity = augment_peak_removal_intensity
         self.augment_intensity = augment_intensity
 
     def __len__(self):
         """Denotes the number of batches per epoch"""
         # TODO: this means we don't see all data every epoch, because the last half-empty batch
-        #  is omitted. I guess that is expected behavior?
-        return int(self.num_turns) * int(np.floor(len(self.training_ids) / self.batch_size))
+        #  is omitted. I guess that is expected behavior? --> Yes, with the shuffling in each epoch that seem OK to me (and makes the code easier).
```
svenvanderburg, line 105: Cool :) I guess you can remove the comments now? Or maybe add a small note to the docstring.

ms2deepscore/data_generators.py
```diff
@@ -10,150 +11,188 @@
...
-            TODO: In same order as what?
-            Array of reference similarity scores (=labels).
-        inchikey_mapping
-            TODO: Needs documentation
+        score_array
+            2D-array of reference similarity scores (=labels).
+            Must be symmetric (score_array[i,j] == score_array[j,i]) and in the
+            same order as inchikey_score_mapping.
+        inchikey_ids
+            List of IDs from unique_inchikeys to use for training.
```
svenvanderburg, line 43: `unique_inchickeys` is `inchikey_score_mapping`?

tests/test_data_generators.py
```diff
@@ -0,0 +1,122 @@
+import os
```
svenvanderburg, line 1: Cool! I like these tests 👌  They are simple, but they touch a lot of the code which is what you care about in this stage.

svenvanderburg (approved):
Really nice changes! Adding the documentation and especially renaming the different ids makes things a lot easier to understand and thus more sustainable! 👍 

There are some PEP8 violations (I spotted a few in the tests), you could add a linter to the CI pipeline, or you could not care :)

tests/test_data_generators.py
florian-huber: Nice, thanks.

ms2deepscore/data_generators.py
florian-huber: Yes you're right I think. Thank.s

ms2deepscore/data_generators.py
florian-huber: Indeed. Will remove that comment.

ms2deepscore/data_generators.py
florian-huber: good catch. thanks!

tests/test_data_generators.py
florian-huber: Sorry, I'll skip that for now...

PR #14, opened 2020-12-22
add BinnedSpectrum class

florian-huber:
- create `BinnedSpectrum` class
- simplify `SpectrumBinner` (remove generator argument storage), and switch to scikit-API style --> `.fit_transform()` and `transform()`
- refactor data generators (now use lists of `BinnedSpectrum` objects as input + some more method inheritances)

### DIFF
--- a/ms2deepscore/BinnedSpectrum.py
+++ b/ms2deepscore/BinnedSpectrum.py
@@ -1,0 +1,39 @@
+class BinnedSpectrum:
+    """Binned spectrum for use with MS2DeepScore."""
+    def __init__(self, binned_peaks: dict, metadata: dict):
+        self.peaks = binned_peaks
+        self._metadata = metadata
+
+    def __eq__(self, other):
+        return \
+            self.peaks == other.peaks and \
+            self.metadata == other.metadata
+
+    def get(self, key: str, default=None):
+        """Retrieve value from :attr:`metadata` dict. Shorthand for
+
+        .. code-block:: python
+
+            val = self.metadata.get("key", default)
+
+        """
+        return self._metadata.get(key, default)
+
+    def set(self, key: str, value):
+        """Set value in :attr:`metadata` dict. Shorthand for
+
+        .. code-block:: python
+
+            self.metadata[key] = val
+
+        """
+        self._metadata[key] = value
+        return self
+
+    @property
+    def metadata(self):
+        return self._metadata.copy()
+
+    @metadata.setter
+    def metadata(self, value):
+        self._metadata = value
--- a/ms2deepscore/BinnedSpectrum.py
+++ b/ms2deepscore/BinnedSpectrum.py
@@ -1,7 +1,7 @@
 class BinnedSpectrum:
     """Binned spectrum for use with MS2DeepScore."""
     def __init__(self, binned_peaks: dict, metadata: dict):
-        self.peaks = binned_peaks
+        self.binned_peaks = binned_peaks
         self._metadata = metadata
 
     def __eq__(self, other):
--- a/ms2deepscore/BinnedSpectrum.py
+++ b/ms2deepscore/BinnedSpectrum.py
@@ -1,6 +1,15 @@
 class BinnedSpectrum:
     """Binned spectrum for use with MS2DeepScore."""
     def __init__(self, binned_peaks: dict, metadata: dict):
+        """
+
+        Parameters
+        ----------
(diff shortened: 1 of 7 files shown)


ms2deepscore/BinnedSpectrum.py
```diff
@@ -0,0 +1,39 @@
+class BinnedSpectrum:
+    """Binned spectrum for use with MS2DeepScore."""
+    def __init__(self, binned_peaks: dict, metadata: dict):
+        self.peaks = binned_peaks
+        self._metadata = metadata
```
svenvanderburg, line 5: As I mentioned before, I would prefer having attributes of the `BinnedSpectrum` class over the `metadata` dict. It is unclear what goes into metadata, what should go in there and what can I expect to be in there. It creates unnecessary nestedness. It makes the code more complex because you need the getters and setters. You get some flexibility back for it, but I am not sure if you really need that.

ms2deepscore/BinnedSpectrum.py
```diff
@@ -0,0 +1,39 @@
+class BinnedSpectrum:
+    """Binned spectrum for use with MS2DeepScore."""
+    def __init__(self, binned_peaks: dict, metadata: dict):
+        self.peaks = binned_peaks
```
svenvanderburg, line 4:
```suggestion
    def __init__(self, binned_peaks: dict, metadata: dict):
        self.binned_peaks = binned_peaks
```

svenvanderburg, line 4: It would be nice to document what binned_peaks looks like.

tests/test_BinnedSpectrum.py
```diff
@@ -0,0 +1,29 @@
...
+    binned_spectrum = BinnedSpectrum(binned_peaks=peaks,
+                                     metadata=metadata)
+    assert binned_spectrum.peaks["20"] == 0.05
+    assert list(binned_spectrum.peaks.keys()) == ['20', '21', '25', '30', '100']
+    assert binned_spectrum.get("inchikey") == "test"
+    assert binned_spectrum.get("smiles") is None
+
+
+def test_BinnedSpectrum_deepcopy():
+    """Test if metadata is made a correct deepcopy"""
```
svenvanderburg, line 21: You are doing a shallow copy in the code I think. I don't know a lot about this, but I suspect that if instead of a string you try mutating a more complex object that this test fails.

ms2deepscore/BinnedSpectrum.py
svenvanderburg: Or the other way around (using `peaks` for both)

svenvanderburg (approved):
Nice improvements! 👍 I left some comments, feel free to ignore!

ms2deepscore/BinnedSpectrum.py
```diff
@@ -0,0 +1,39 @@
+class BinnedSpectrum:
```
svenvanderburg, line 1: modules in python are `snake_case` not `CamelCase`

ms2deepscore/BinnedSpectrum.py
florian-huber: You are right that attributes would be cleaner here. In particular now that we hardly need a lot of metadata entries yet. But I have chosen this for a start because it is analogue to `Spec2Vec` and `matchms` which might have some advantages when using those things together. I will hence leave it for now, but let's check on this later.

ms2deepscore/BinnedSpectrum.py
florian-huber: Good point. I'll change it elsewhere, too.

tests/test_BinnedSpectrum.py
florian-huber: I thought I am doing a deepcopy already. Can you think of an example that should fail?

ms2deepscore/BinnedSpectrum.py
florian-huber: That's right, but classes are CamelCase. Or did you mean another part?

ms2deepscore/BinnedSpectrum.py
florian-huber: Added a sentence to the docstring.

florian-huber:
Thanks a lot for reviewing @svenvanderburg. (I left a few things open, but we can still work on that in another PR!)

PR #22, opened 2021-01-13
adapt test to changed generator

florian-huber:
Merging PR #16  made build fail because generators were changed while PR was prepared. Will be fixed here.

- some parts of the generators were *missing* (method to adapt settings and outdated use of `self.num_turns`)


florian-huber (commented):
just self-reviewed...

PR #25, opened 2021-01-14
Use a dataframe for labels in data generators

svenvanderburg:
Instead of `score_array` and `inchikey_score_mapping` we now pass a pandas dataframe `labels_df` with all the reference scores for inchikey pairs.


ms2deepscore/data_generators.py
```diff
@@ -171,46 +166,52 @@ def on_epoch_end(self):
...
 
-        # Part 2 - if still no match is found take the 2nd-highest score ID
-        if len(idx) == 0:
-            second_highest_id = self.score_array[inchikey_id1, np.array([x for x in self.inchikey_ids if x != inchikey_id1])].argmax()
-            if second_highest_id > inchikey_id1:
-                second_highest_id += 1
-            inchikey_id2 = self.inchikey_ids[second_highest_id]
+        # Part 2 - if still no match is found take the inchikey that has highest similarity score
+        # with inchikey1
+        # TODO: Why are we taking the highest score here? Shouldn't we take the one that is closest
```
svenvanderburg, line 209: @florian-huber what do you think?

ms2deepscore/data_generators.py
```diff
@@ -382,71 +376,44 @@ def __len__(self):
...
+        return labels_df.loc[selected_inchikeys, selected_inchikeys]
+
     def on_epoch_end(self):
         """Updates indexes after each epoch"""
-        self.indexes = np.tile(np.arange(len(self.inchikey_ids)), int(self.settings["num_turns"]))
+        self.indexes = np.tile(np.arange(len(self.labels_df)), int(self.settings["num_turns"]))
         if self.settings["shuffle"] == True:
             np.random.shuffle(self.indexes)
 
-    def __find_match_in_range(self, inchikey_id1, target_score_range, max_range=0.4):
```
svenvanderburg, line 416: I looked very carefully, but I couldn't find any differences with `DataGeneratorAllSpectrums.__find_match_in_range`. That's correct right?

ms2deepscore/data_generators.py
```diff
@@ -245,40 +246,37 @@ def __data_generation(self, spectrum_inchikey_ids_batch):
...
-        inchikey_ids
-            List of IDs from unique_inchikeys to use for training.
-        inchikey_score_mapping
-            Array of all unique inchikeys. Must be in the same order as the scores
-            in scores_array.
+        labels_df
+            Pandas DataFrame with reference similarity scores (=labels) for compounds identified
+            by inchikeys. Columns and index should be inchikeys, the value in a row x column
+            depicting the similarity score for that pair. Must be symmetric
+            (labels_df[i,j] == labels_df[j,i]) and column names should be identical to the index.
```
svenvanderburg, line 277: Is this clear enough?

ms2deepscore/data_generators.py
florian-huber: Yes, I think that's correct. I was also thinking of creating a parent class and inherit from there... but then postponed it for now.

ms2deepscore/data_generators.py
```diff
@@ -17,22 +17,19 @@ class DataGeneratorAllSpectrums(Sequence):
     other spectrum that corresponds to a reference score as defined in same_prob_bins.
     """
     def __init__(self, spectrums_binned: List[BinnedSpectrum], spectrum_ids: list,
-                 score_array: np.ndarray, inchikey_score_mapping: np.ndarray,
-                 dim: int, **settings):
+                 labels_df: pd.DataFrame, dim: int, **settings):
```
florian-huber, line 20: Could be only me, but I first stumbled a bit about the name `labels`. It of course is totally standard from the ML community side, but still might not be 100% obvious here? Maybe `reference_scores_df`? But feel free to leave it in.

ms2deepscore/data_generators.py
```diff
@@ -61,25 +58,24 @@ def __init__(self, spectrums_binned: List[BinnedSpectrum], spectrum_ids: list,
...
         self.inchikeys_all = np.array([x.get("inchikey") for x in spectrums_binned])
         # TODO: add check if all inchikeys are present (should fail for missing ones)
         self.dim = dim
 
         self.on_epoch_end()
 
+    @staticmethod
+    def _validate_labels(labels_df: pd.DataFrame):
+        if set(labels_df.index) != set(labels_df.columns):
+            raise ValueError(f'index and columns of labels_df are not identical')
```
florian-huber, line 77: Nice check!

ms2deepscore/data_generators.py
```diff
@@ -171,46 +166,52 @@ def on_epoch_end(self):
...
-    def __find_match_in_range(self, inchikey_id1, target_score_range, max_range=0.4):
+    @staticmethod
+    def _exclude_nans_from_labels(labels_df: pd.DataFrame):
+        """Exclude nans in labels_df, exclude columns and rows if there is any NaN
+        value"""
+        clean_df = labels_df.dropna(axis='rows')  # drop rows with any NaN
+        clean_df = clean_df[clean_df.index]  # drop corresponding columns
+        n_dropped = len(labels_df) - len(clean_df)
+        if n_dropped > 0:
+            print(f"{n_dropped} nans among {len(labels_df)} labels will be excluded.")
```
florian-huber, line 177: That's much nicer than it was before I think.

ms2deepscore/data_generators.py
florian-huber: My thinking behind taking the highest non-identical inchikey:
It sometimes happens that there is no Inchikey for a selected score-bin. Due to the strong bias in the data towards low similarity scores this only occurs for the higher bins (say >0.5). A perfect balancing of the data is hence not possible by the method provided here. The extending the range part ("Part 1") should compensate most of this. I didn't spend much time on this and simply felt like biasing the choice in "Part 2" towards highest-possible pairs could help as well...  but your solution seems a bit cleaner. So I am happy keeping your changes here (I expect that in practice this should often come down to the same choice).

ms2deepscore/data_generators.py
florian-huber: Yes, sounds good. (I commented on the term "labels" above, but I'm already getting used to it ;) )

florian-huber (approved):
Super nice!
I think that makes the code and API much cleaner (and more robust I would guess).

ms2deepscore/data_generators.py
svenvanderburg: I've opened #32 for this.

ms2deepscore/data_generators.py
svenvanderburg: OK, I think if you stumble over it then others might as well. I renamed it to `reference_scores_df` like you suggest. :)

PR #28, opened 2021-01-14
Refactor data generators

svenvanderburg:
Changes:
* Renamed `spectrums_binned` to `binned_spectrums` so it matches `BinnedSpectrum`
* Get rid of `inchikeys_all`, instead use the `inchikey` in `BinnedSpectrum` instances in `binned_spectrums`
* Factor out `_get_spectrum_with_inchikey`
* Define `SpectrumPair` namedtuple, representing a pair of `BinnedSpectrum` instances.
* Move `__data_generation` to base class, it now takes a generator of `SpectrumPair` objects that should be implemented in the inheriting classes.
* Move `__get_item__` method to base class
* Implement `_spectrum_pair_generator` based on old `__get_item__` methods


florian-huber (approved):
Really very nice code refactoring!

svenvanderburg:
> Really very nice code refactoring!
> I agree with all changes made (I was first a bit surprised by the `SpectrumPair`, but it does make the generator code easier to read I think).

Yeah, it seems a bit verbose, but it makes it explicit what goes in and out of a few methods.

Issue #24, opened 2021-01-14
DataGeneratorAllSpectrums pair selection mechanism leaks


svenvanderburg:
For `DataGeneratorAllSpectrums` the 'selection' of the data is based on spectrums (i.e. user passes a list of spectrum_ids to generate data from). As a user I would expect that for each of the spectrum pairs that are generated, both spectrums are in this list that I supplied (because I only want for a limited set of spectrums to be in my training vs test data for example). 

However, only the 'first' spectrum in a pair is in this list of spectrum_ids, whereas the 'second' spectrum can be outside of the spectrum_ids list, see `DataGeneratorAllSpectrums.__find_match_in_range`:
https://github.com/matchms/ms2deepscore/blob/0d80b9c0f5b5add334882aa9ab36da714a7a183c/ms2deepscore/data_generators.py#L181-L213

This is bad, because it could mean that training data leaks into test data.

svenvanderburg:
@florian-huber did I miss something here or is this indeed a problem?

florian-huber:
@svenvanderburg I think you are right. Good catch!
I thought I had addressed that somewhere, but it looks like I didn't. So I will take care of this today!
(probably I will simply remove all inchikeys that are not among the selected spectrums and then only pick from the selected spectrums once `inchikey_id2` is chosen)

PR #30, opened 2021-01-15
Fix generator leak

florian-huber:
Addresses #24  
Changes to `DataGeneratorAllSpectrums`:
- removed `spectrum_ids`
- added `_exclude_not_selected_inchikeys` method


tests/test_data_generators.py
```diff
@@ -79,6 +79,76 @@ def test_DataGeneratorAllSpectrums():
...
+                          'JFFHVIUZNPTGGR-WJLGXSQGSA-N',
+                          'JGCSKOVQDXEQHI-UHFFFAOYSA-N',
+                          'VCBNPTWPJQLHQN-NYAJDEOCSA-N',
+                          'ZBAMSLOMNLECFR-IEAZIUSSSA-N']
+    found_inchikeys = test_generator.labels_df.columns.to_list()
+    found_inchikeys.sort()
+    assert found_inchikeys == expected_inchikeys, \
+        "Expected different InChIKeys to remain in labels_df"
+
+    # Test if ibkt exoected labels are returned by generator
```
svenvanderburg, line 113: ibkt exoected?

ms2deepscore/data_generators.py
```diff
@@ -333,6 +335,17 @@ def _exclude_nans_from_labels(labels_df: pd.DataFrame):
             print(f"{n_dropped} nans among {len(labels_df)} labels will be excluded.")
         return clean_df
 
+    def _exclude_not_selected_inchikeys(self):
```
svenvanderburg, line 338: It is not immediately clear from the method name and the docstring from *what* the inchikeys are removed (i.e. labels_df)

svenvanderburg (commented):
Nice solution 😄  (I am curious how this affects the results!)

I have left some comments, but I also have a more major suggestion:

If we make `_exclude_not_selected_inchikeys` also exclude the inchikeys also from `binned_spectrums`, I think we can safely leave the `_get_spectrum_with_inchikey` method unaltered in the base class.

Thinking even further it would also be possible to get rid of `spectrum_ids` argument for `DataGeneratorAllSpectrums`, instead one could just pass the list of `binned_spectrums` that one wants to select from!

I have a preference for that last option.

tests/test_data_generators.py
florian-huber: looks like: random keyboard typing while editor still open (multiple screen problem)

florian-huber:
Thanks for the comments @svenvanderburg . I guess I made it unnecessarily complicated 😬 ...
I will remove the `spectrum_ids` which will already solve most of the leaking possibilities.

ms2deepscore/data_generators.py
florian-huber: True. I added `labels_df` as argument and extended the docstring.

tests/test_data_generators.py
florian-huber: Fixed.

svenvanderburg (approved):
Nice! This really cleans up things :)

florian-huber:

PR #36, opened 2021-01-20
Model import and export

florian-huber:
OK, this got a bit bigger than I thought. In the end this are the changes:
- fixed the binning functions itself to properly handle the `mz_max` argument (ignoring peaks > mz_max)
- Included `.spectrum_binner` in `SiameseModel` class
- added `.save()` method --> stores keras model + spectrum_binner to one `hdf5` file.
- added `load_model` function
- changed import style and added `BinnedSpectrumType` (because half way I ran into some annoying TypeError problems)
- added tests for saving and loading a model.


ms2deepscore/models/SiameseModel.py
```diff
@@ -31,12 +61,16 @@ def __init__(self,
         dropout_rate
             Dropout rate to be used in the base model
         """
-        self.base = self._get_base_model(input_dim=input_dim,
+        assert spectrum_binner.known_bins is not None, \
+            "spectrum_binner does not contain known bins (run .fit_transform() on training data first!)"
+        self.spectrum_binner = spectrum_binner
+        self.input_dim = len(spectrum_binner.known_bins)
```
svenvanderburg, line 67: Cool, nice that we can get this from `spectrum_binner`!

ms2deepscore/models/load_model.py
```diff
@@ -0,0 +1,54 @@
...
+        keras_model = hdf5_format.load_model_from_hdf5(f)
+
+    # Reconstitute spectrum_binner
+    binner_dict = json.loads(binner_json)
+    spectrum_binner = SpectrumBinner(binner_dict["number_of_bins"],
+                                     binner_dict["mz_max"], binner_dict["mz_min"],
+                                     binner_dict["peak_scaling"],
+                                     binner_dict["allowed_missing_percentage"])
+    spectrum_binner.peak_to_position = {int(key): value for key, value in binner_dict["peak_to_position"].items()}
+    spectrum_binner.known_bins = binner_dict["known_bins"]
```
svenvanderburg, line 39: Maybe this deserialization code should live in the same module as `SpectrumBinner`? (In case you want to reuse it somehwere) Same goes for the serialisation code in `SiameseModel.save()`. You could for example have a `from_json()` class method and `to_json()` method for `SpectrumBinner` class?

ms2deepscore/models/load_model.py
```diff
@@ -0,0 +1,54 @@
...
+        Filename. Expecting saved SiameseModel.
+
+    """
+    with h5py.File(filename, mode='r') as f:
+        binner_json = f.attrs['spectrum_binner']
+        keras_model = hdf5_format.load_model_from_hdf5(f)
+
+    # Reconstitute spectrum_binner
+    binner_dict = json.loads(binner_json)
+    spectrum_binner = SpectrumBinner(binner_dict["number_of_bins"],
```
svenvanderburg, line 34: Would `SpectrumBinner(**binner_dict)` not work?

ms2deepscore/models/load_model.py
```diff
@@ -0,0 +1,54 @@
...
+    spectrum_binner.known_bins = binner_dict["known_bins"]
+
+    # Extract parameters for SiameseModel
+    embedding_dim = keras_model.layers[2].output_shape[1]
+    base_dims = []
+    for layer in keras_model.layers[2].layers:
+        if "dense" in layer.name:
+            base_dims.append(layer.output_shape[1])
+        elif "dropout" in layer.name:
+            dropout_rate = layer.rate
```
svenvanderburg, line 48: Would it make sense to move this to `SiameseModel`? Maybe add a class method: `from_keras_model()`?

tests/test_models.py
```diff
@@ -43,3 +53,53 @@ def test_siamese_model():
...
+    X, y = test_generator.__getitem__(0)
+    embeddings = model.base.predict(X[0])
+    assert isinstance(embeddings, np.ndarray), "Expected numpy array"
+    assert embeddings.shape[0] == test_generator.settings["batch_size"] == 32, \
+        "Expected different batch size"
+    assert embeddings.shape[1] == model.base.output_shape[1] == 200, \
+        "Expected different embedding size"
+
+
+def test_save_and_load_jmodel(tmp_path):
```
svenvanderburg, line 80:
```suggestion
def test_save_and_load_model(tmp_path):
```

svenvanderburg (approved):
Looks good 👍 ! I left a few mostly code-organisation comments, do with it that you think's best!

ms2deepscore/models/load_model.py
florian-huber: Unfortunately that doesn't work. the `binner_dict` contains the entire `SpectrumBinner`, but only few things are expected in the first init step. (we can probably make this nicer at some point but I will leave it as is for now)

ms2deepscore/models/load_model.py
florian-huber: Good ideas!
I added a `to_json()` method.
Not sure about the `from_json()` method though. How would you implement that? I mean, how would you deal with the __init__ part then?

florian-huber: As an alternative I could move the part you marked here into its own helper function in `load_model.py`.

ms2deepscore/models/load_model.py
florian-huber: This is a good point and similar to the one about importing a SpectrumBinner from json. In both cases I feel that we have to adapt the __init__() method first. I'll look into it a bit.

florian-huber: I now indeed moved it to `SiameseModel` in a way that makes this piece of code obsolete. 
Maybe you can have a quick look at it again @svenvanderburg?

ms2deepscore/models/load_model.py
florian-huber: For now I left it in `load_model` but organized it a bit clearer in a sub-function.

ms2deepscore/models/load_model.py
svenvanderburg: That looks good! :)

ms2deepscore/models/load_model.py
svenvanderburg: I totally like what you did now, so you can leave it like that (although I would vote for the `binner_from_json` function to move to the spectrumbinner module so the deserialisation code lives close to the class).

Just to show you how you could do this using a classmethod:
```python
class SpectrumBinner:
    def __init__():
         # The normal constructor
    @classmethod
    def from_json(json_str: str):
        binner_dict = json.loads(binner_json)
        spectrum_binner = cls(binner_dict["number_of_bins"],
                                         binner_dict["mz_max"], binner_dict["mz_min"],
                                         binner_dict["peak_scaling"],
                                         binner_dict["allowed_missing_percentage"])
        spectrum_binner.peak_to_position = {int(key): value for key, value in binner_dict["peak_to_position"].items()}
        spectrum_binner.known_bins = binner_dict["known_bins"]
        return spectrum_binner
        
spectrum_binner = SpectrumBinner(*args, **kwargs)  # Normal construction
spectrum_binner = SpectrumBinner.from_json(json_str) # Construction from json
```

ms2deepscore/models/load_model.py
florian-huber: Great, thanks!

ms2deepscore/models/load_model.py
florian-huber: That's indeed nicer! So I switched to that classmethod.

florian-huber:

PR #49, opened 2021-02-05
Switch to inchikey14

florian-huber:


florian-huber:
HI @svenvanderburg 
With this changed, the code seems fully functional! (currently running on actual data)
I would hence move forward and add it to pypi once this is merged, or what would you say?

ms2deepscore/data_generators.py
```diff
@@ -228,7 +237,7 @@ def __data_generation(self, spectrum_pairs: Iterator[SpectrumPair]):
             for i_spectrum, spectrum in enumerate(pair):
                 idx, values = self._data_augmentation(spectrum.binned_peaks)
                 X[i_spectrum][i_pair, idx] = values
-            y[i_pair] = self.reference_scores_df[pair[0].get('inchikey')][pair[1].get('inchikey')]
+            y[i_pair] = self.reference_scores_df[pair[0].get("inchikey")[:14]][pair[1].get("inchikey")[:14]]
```
svenvanderburg, line 240: Now at every inchikey-based retrieval from `reference_scores_df` or `binned_spectrums` (this line is an example), we transform inchikey to inchikey14. Can we not transform the inchikeys once for both the reference_scores_df and binned_spectrums when we initialize the class?

svenvanderburg: And then cleanest would be to do it in a method `transform_data_to_inchikey14` or something.

svenvanderburg (commented):
Nice that you spotted the heavy list comprehension! It's a shame that we now have to introduce again another `id` vector, but I guess we have to. 

I guess the reason for using inchikey14 is sane, but just to double check: is it? And can we expect collisions to happen if a user for some reason uses longer inchikeys and we without him knowing transform to shorter keys?

ms2deepscore/data_generators.py
```diff
@@ -69,8 +69,17 @@ def __init__(self, binned_spectrums: List[BinnedSpectrumType],
...
         self.reference_scores_df = self._exclude_nans_from_labels(reference_scores_df)
-        # TODO: add check if all inchikeys are present (should fail for missing ones)
+        self._collect_inchikeys()
         self.dim = dim
+        
+    def _collect_inchikeys(self):
+        """Collect all inchikeys14 (first 14 characters) of all binned_spectrums
+        and check if all are present in the reference scores as well.
+        """
+        self.inchikeys = np.array([s.get("inchikey")[:14] for s in self.binned_spectrums])
```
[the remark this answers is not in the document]
florian-huber: Good point. Agree with the change.

ms2deepscore/data_generators.py
florian-huber: I will add that change to the `reference_score_df`, but not to the `binned_spectrums` because the later contains a shallow copy pointing to the original spectrums (and for other purposes the full inchikey might be needed). 
(I agree that the `[:14]` additions are a bit ugly, but I would keep it for now, I think)

justinjjvanderhooft: Indeed, it's good to keep the original Inchikeys as well!

ms2deepscore/data_generators.py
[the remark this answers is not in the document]
florian-huber: Done.

florian-huber:
Thanks for the review @svenvanderburg 

> can we expect collisions to happen if a user for some reason uses longer inchikeys and we without him knowing transform to shorter keys?

That's true. With the changes to your above comments the code will also work if a user imports reference scores with full inchikey indices, but those will be truncated in the data generator. So two entries with different stereochemistry will become indistinguishable. (MS2DeepScore doesn't have any user we don't know of yet... so that should be fine I guess 😄 )
I is now in the docstring documentation, but we could add a warning if full inchikeys are found via logging (I'll add a new issue).

PR #55, opened 2021-02-08
Add noise to generator and update model architecture

florian-huber:
- added noise peaks to data generator augmentation part
- added L1 and L2 regularization to first dense network layer


florian-huber:
Rather small changes. Will merge them without extra review.

PR #58, opened 2021-02-10
add new confusion matrix plotting function

florian-huber:
- Add another plotting function (similar to confusion matrix)


florian-huber (commented):
Just did a simple self-review here, because this is mostly about adding a plotting function we later anyway might remove/adapt/expand on.

Issue #63, opened 2021-03-01
make SiameseModel architecture more flexible


florian-huber:
Now we only allow 4 dense layers, but that could easily be changed to `n` layers by simply giving a list of layer dimensions as input.

florian-huber:
Implemented in #64

PR #64, opened 2021-03-01
allow different number of layers

florian-huber:
Adsressing #63 

- Allow user to define `SiameseModel` architecture more freely, including the number of layers

### DIFF
--- a/ms2deepscore/models/SiameseModel.py
+++ b/ms2deepscore/models/SiameseModel.py
@@ -1,5 +1,5 @@
 from pathlib import Path
-from typing import List, Union
+from typing import Tuple, Union
 import h5py
 from tensorflow import keras
 from tensorflow.python.keras.saving import hdf5_format
@@ -54,7 +54,7 @@ class SiameseModel:
         spectrum_binner
             SpectrumBinner which is used to bin the spectra data for the model training.
         base_dims
-            List of integers depicting the dimensions of the desired hidden
+            Tuple of integers depicting the dimensions of the desired hidden
             layers of the base model
         embedding_dim
             Dimension of the embedding (i.e. the output of the base model)
--- a/ms2deepscore/models/SiameseModel.py
+++ b/ms2deepscore/models/SiameseModel.py
@@ -104,18 +104,18 @@ class SiameseModel:
         model_input = keras.layers.Input(shape=input_dim, name='base_input')
         for i, dim in enumerate(dims):
             if i == 0:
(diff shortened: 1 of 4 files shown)


ms2deepscore/models/SiameseModel.py
```diff
@@ -98,20 +98,22 @@ def save(self, filename: Union[str, Path]):
...
+                   model_input)
+            else:
+                embedding = keras.layers.Dense(dim, activation='relu', name='dense'+str(i+1),
+                                               kernel_regularizer=keras.regularizers.l1_l2(l1=1e-6, l2=1e-6))(
+                   embedding)
+            embedding = keras.layers.BatchNormalization(name='normalization'+str(i+1))(embedding)
+            embedding = keras.layers.Dropout(dropout_rate, name='dropout'+str(i+1))(embedding)
+
         embedding = keras.layers.Dense(embedding_dim, activation='relu', name='embedding')(
             embedding)
```
svenvanderburg, line 118: Only the last layer I would call 'embedding', maybe `x` seems more appropriate to assign the intermediate layers to. Also the if/else statement is not necessary I think.
```suggestion
        x = model_input
        for i, dim in enumerate(dims):
                x = keras.layers.Dense(dim, activation='relu', name='dense'+str(i+1),
                                               kernel_regularizer=keras.regularizers.l1_l2(l1=1e-6, l2=1e-6))(
                   x)
            x = keras.layers.BatchNormalization(name='normalization'+str(i+1))(x)
            x = keras.layers.Dropout(dropout_rate, name='dropout'+str(i+1))(x)

        embedding = keras.layers.Dense(embedding_dim, activation='relu', name='embedding')(x)
```

svenvanderburg (approved):
Nice new feature! I just left 1 suggestion to improve the code a bit :)

ms2deepscore/models/SiameseModel.py
florian-huber: I also wasn't fully happy with the if-else, but the input layer needed a different name for the final construction of the `Model` in `return keras.Model(model_input, embedding, name='base')`.

florian-huber: But changing all other `embedding` to `x` is fine for me (the linter might complain because it doesn't like 1-character variables, but I can check that).

florian-huber:
I now used `model_layer` instead of `x`, and I also realized that I will further work on this anyway in #65

Issue #53, opened 2021-02-08
Implement Monte-Carlo Dropout (uncertainty evaluation)


florian-huber:
The `SiameseModel` is trained using dropout, which helps to prevent overfitting. But it also is there to help assess the uncertainty of the model by using Monte-Carlo Dropout. 

Right now this lives only in a few notebooks, so it should become part of the main code.

florian-huber:
One trade-off issue here is the following:
Performance-wise, the main computational cost of our scores are the 2 first steps: `(1) Spectrum -> BinnedSpectrum`, and `(2) BinnedSpectrum -> embedding`. So ideally we do not want to duplicate such steps.
However, if we compute scores between a high number of spectrums that can get complicate memory-wise. Say, we compute the scores for all pairs between 10,000 spectrums, than we need to compute 10,000 embeddings and run `cosine_similarity_matrix`. That's actually still fine.
But what if we increase the number further, and/or increase `n_ensemble` (code example above) so that we compute `n_ensemble` different embeddings for every spectrum...

One alternative is to break it up into sub-parts, but then we have to calculate some embeddings several times (hence add computational cost).

Another way to go might be, to first compute the most expensive part that can still be used for all further steps for all spectrums.
- The first step `(1) Spectrum -> BinnedSpectrum` can easily be done for all spectrums.

- The second step could be splitted: `(2a) BinnedSpectrum -> 1.layer embedding` `(2b) 1.layer embedding -> output embedding`    
The reason for this would be that the 1st dense layer does not use `dropout` so it will always give the same result and at the same time compress the input data quiet a bit!

florian-huber:
Spec2Vec has been added to GNPS (release 27)!
--> https://github.com/CCMS-UCSD/GNPS_Workflows/releases

florian-huber:
Done with #65

PR #65, opened 2021-03-05
Add monte carlo dropout

florian-huber:
First implementation to address #53 .
That might not yet be the most elegant way to implement things, but it could serve as a solid start.
- added `MS2DeepScoreMonteCarlo`
- added helper functions to compute the means and standard deviations

[edit] minor additional changes through merge with main (due to #66 )


ms2deepscore/models/SiameseModel.py
```diff
@@ -105,34 +106,60 @@ def save(self, filename: Union[str, Path]):
...
+                       dropout_rate: float = 0.25,
+                       l1_reg: float = 1e-6,
+                       l2_reg: float = 1e-6,
+                       dropout_always_on: bool = False) -> keras.Model:
+        """Create base model for Siamaese network.
+
+        Parameters
+        ----------
+        input_dim : int
+            DESCRIPTION.
```
svenvanderburg, line 121: I think this should be updated.

ms2deepscore/models/SiameseModel.py
```diff
@@ -105,34 +106,60 @@ def save(self, filename: Union[str, Path]):
...
+        base_dims
```
svenvanderburg, line 122: There is no `base_dims`, just `dims`

ms2deepscore/models/SiameseModel.py
```diff
@@ -105,34 +106,60 @@ def save(self, filename: Union[str, Path]):
...
-                                                 name='dense'+str(i+1))(model_layer)
-            model_layer = keras.layers.BatchNormalization(name='normalization'+str(i+1))(model_layer)
-            model_layer = keras.layers.Dropout(dropout_rate, name='dropout'+str(i+1))(model_layer)
+                model_layer = Dense(dim, activation='relu', name='dense'+str(i+1))(model_layer)
+            model_layer = BatchNormalization(name='normalization'+str(i+1))(model_layer)
+            if dropout_always_on:
+                model_layer = Dropout(dropout_rate, name='dropout'+str(i+1))(model_layer,
+                                                                             training=True)
+            else:
+                model_layer = Dropout(dropout_rate, name='dropout'+str(i+1))(model_layer)
```
svenvanderburg, line 153: Could be simplified to:
```suggestion

            model_layer = Dropout(dropout_rate, name='dropout'+str(i+1))(model_layer,
                                                                             training=dropout_always_on)
```

ms2deepscore/MS2DeepScoreMonteCarlo.py
```diff
@@ -0,0 +1,171 @@
...
+from .typing import BinnedSpectrumType
+
+
+class MS2DeepScoreMonteCarlo(BaseSimilarity):
+    """Calculate MS2DeepScore ensemble similarity scores and STD between a reference
+    and a query using Monte-Carlo Dropout
+
+    Using a trained model, binned spectrums will be converted into spectrum
+    vectors using a deep neural network. The MS2DeepScoreMonteCarlo similarity is then
+    the mean of n_ensemble x n_ensemble cosine similarity score between two spectrum
```
svenvanderburg, line 17: Is'nt it the man of n_ensemble cosine similarity scores? Instead of a matrix of n_ensemble x n_ensemble?

ms2deepscore/MS2DeepScoreMonteCarlo.py
```diff
@@ -0,0 +1,171 @@
...
+    def _create_input_vector(self, binned_spectrum: BinnedSpectrumType):
+        """Creates input vector for model.base based on binned peaks and intensities"""
+        X = np.zeros((1, self.input_vector_dim))
+
+        idx = np.array([int(x) for x in binned_spectrum.binned_peaks.keys()])
+        values = np.array(list(binned_spectrum.binned_peaks.values()))
+        X[0, idx] = values
+        return X
+
+    def _create_monte_carlo_encoder(self):
```
svenvanderburg, line 76: I would stick to `base` instead of `encoder`

ms2deepscore/MS2DeepScoreMonteCarlo.py
```diff
@@ -0,0 +1,171 @@
...
+        Parameters
+        ----------
+        reference:
+            Reference spectrum.
+        query:
+            Query spectrum.
+
+        Returns
+        -------
+        ms2ds_similarity
```
svenvanderburg, line 107: It returns the mean + std

ms2deepscore/MS2DeepScoreMonteCarlo.py
```diff
@@ -0,0 +1,171 @@
...
+            Reference spectrum.
+        queries:
+            Query spectrum.
+        is_symmetric:
+            Set to True if references == queries to speed up calculation about 2x.
+            Uses the fact that in this case score[i, j] = score[j, i]. Default is False.
+
+        Returns
+        -------
+        ms2ds_similarity
```
svenvanderburg, line 134: Returns matrix of mean similarity scores and matrix of std's

ms2deepscore/MS2DeepScoreMonteCarlo.py
```diff
@@ -0,0 +1,171 @@
...
+        progress_bar:
+            Set to True to monitor the embedding creating with a progress bar.
+            Default is False.
+        """
+        self.model = model
+        self.n_ensembles = n_ensembles
+        self.input_vector_dim = self.model.base.input_shape[1]  # TODO: later maybe also check against SpectrumBinner
+        self.output_vector_dim = self.model.base.output_shape[1]
+        self.progress_bar = progress_bar
+        self.partial_model = self._create_monte_carlo_encoder()
```
svenvanderburg, line 65: Maybe be consistent in naming. I would suggest:
```python
self.monte_carlo_base = self._create_monte_carlo_base()
```

svenvanderburg (approved):
Looks good, I just added some small comments!

ms2deepscore/models/SiameseModel.py
florian-huber: That could cause trouble, because it would freeze the model in `training=False` for the default state. In that case dropout would stay off even during actual training.

ms2deepscore/models/SiameseModel.py
florian-huber: Done.

ms2deepscore/models/SiameseModel.py
florian-huber: Done

ms2deepscore/MS2DeepScoreMonteCarlo.py
```diff
@@ -0,0 +1,171 @@
...
+        return X
+
+    def _create_monte_carlo_encoder(self):
+        """Rebuild base network with training=True"""
+        dims = []
+        for layer in self.model.base.layers:
+            if "dense" in layer.name:
+                dims.append(layer.units)
+            if "dropout" in layer.name:
+                dropout_rate = layer.rate
```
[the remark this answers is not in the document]
florian-huber: Thanks, good point. I now added a print statement if different dropout rate values are found.

ms2deepscore/MS2DeepScoreMonteCarlo.py
florian-huber: Done.

ms2deepscore/MS2DeepScoreMonteCarlo.py
florian-huber: Done

ms2deepscore/MS2DeepScoreMonteCarlo.py
florian-huber: done

ms2deepscore/MS2DeepScoreMonteCarlo.py
florian-huber: Added this to the documentation (and now also user can chose between median and mean). I recently thought that median would be better, because it is more robust against outliers, so I made median the default.

ms2deepscore/MS2DeepScoreMonteCarlo.py
florian-huber: It is indeed not totally intuitive... but `n_ensemble` means that `n_ensemble` embeddings are created per spectrum. And since we are interested in pair-wise scores, that means we can do `n_ensemble x n_ensemble` possible comparisons with that.

florian-huber:
I edited the code according to your comments.
I also changed the network architecture a bit to skip the dropout for the first layer (which is what I had done for my local screens anyway).

PR #78, opened 2021-03-24
Only use new peaks for peak addition

svenvanderburg:
Addresses #78


florian-huber (approved):
Nice! Thanks Sven, looking good to me.

PR #79, opened 2021-03-24
Use reference score closest to the bin as backup for find_match_in_range

### LINKED ISSUE #72: Select pair with closest similarity to bin center instead of highest similarity in data generation.
In cases where the similarity label for none of the spectrum pairs fell into the selected bin, we selected the pair that had the highest similarity.I think I've said this before, it doesn't make much sense to pick the highest one, it should be the one closest to the center of the bin.

- [x] Update text in methods section accordingly

svenvanderburg:
This actually seems like a much simpler solution. We just widen the range until we find a suitable pair.

Closes #72


florian-huber (approved):
Nice Sven! Looks much clearer now (and also makes more sense I assume).

PR #83, opened 2021-03-25
fixe position of random seed

florian-huber:


ms2deepscore/data_generators.py
```diff
@@ -179,10 +176,10 @@ def _set_generator_parameters(self, **settings):
                 settings[key] = defaults[key]
         assert 0.0 <= settings["augment_removal_max"] <= 1.0, "Expected value within [0,1]"
         assert 0.0 <= settings["augment_removal_intensity"] <= 1.0, "Expected value within [0,1]"
-        if settings["use_fixed_set"] and settings["num_turns"] != 1:
-            raise ValueError('When using a fixed set, num_turns can only be 1')
         if settings["use_fixed_set"] and settings["shuffle"]:
             warnings.warn('When using a fixed set, data will not be shuffled')
+        if settings["use_fixed_set"]:
```
svenvanderburg, line 181: Did you test that this works? Because I tried and tested this in: #80 and it seemed that it didn't work.

ms2deepscore/data_generators.py
```diff
@@ -179,10 +176,10 @@ def _set_generator_parameters(self, **settings):
                 settings[key] = defaults[key]
         assert 0.0 <= settings["augment_removal_max"] <= 1.0, "Expected value within [0,1]"
         assert 0.0 <= settings["augment_removal_intensity"] <= 1.0, "Expected value within [0,1]"
-        if settings["use_fixed_set"] and settings["num_turns"] != 1:
```
svenvanderburg, line 182: Now that I think of it, this should indeed work (I was worried that the 'caching' mechanism would fail)

svenvanderburg (changes requested):
Allowing `num_turns > 1` makes total sense. I am a bit worried about the random seed change. Let's discuss this afternoon.

ms2deepscore/data_generators.py
svenvanderburg: Discussed: this actually passed the test in #80

PR #84, opened 2021-03-31
Add note on data selection reference scores

svenvanderburg:


florian-huber (approved):
Thanks for checking this Sven!

PR #85, opened 2021-03-31
update distribution plot

florian-huber:


florian-huber:
This is only to slightly make the distributions plot prettier. Will merge it.

PR #86, opened 2021-04-02
expand MS2DeepScoreMonteCarlo

florian-huber:
- make scores a structured numpy array (so it can be accessed by `scores["score"]` and `scores["uncertainty"]`)
- add option to have dropout in first network layer (default is False)
- allow to chose between mean (with uncertainty measured by STD) and median (uncertainty measured by IQR).


svenvanderburg (approved):
Nice additions! Code looks clean and documentation is sufficient :)

florian-huber:

Issue #91, opened 2021-08-17
TensorFlow update breaks SiameseModel._construct_from_keras_model()


dianacarvalho1:
The current TensorFlow version 2.6.0 (released in August 11th) breaks the `SiameseModel._construct_from_keras_model()` method. The assertion `assert isinstance(keras_model, keras.Model)` is not `True` anymore and throws the error `AssertionError: Expected keras model as input.`. 
I would suggest fixing the TensorFlow version to 2.5.0 or even updating the code.

Thank you!

florian-huber:
Thanks @dianacarvalho1 for making this issue!
I will run some quick tests. I prefer updating the code so that it is also compatible with TensorFlow 2.6 and will fix the version if that won't easily work.

florian-huber:
I could fix it for tensorflow 2.6 and will now add tests for 2.4, 2.5 and 2.6

florian-huber:
Fixed with #92 . Will be part of the next release (to be done asap).

dianacarvalho1:
Thank you so much @florian-huber!

PR #92, opened 2021-08-19
Update SiameseModel to tensorflow 2.6

florian-huber:
Some of our code breaks with the most recent version of Tensorflow (see issue #91).

- I had to change the assertion that looks for a valid keras model. 
(we could maybe add another assert such as `assert model.layers, "Expected valid keras model as input"` ?)
- To be sure it doesn't now break for former tensorflow version I added CI tests for Tensorflow `>=2.4,<2.5` and `>=2.5, <2.6`.
- Tensorflow recommends to not use `tensorflow.python.keras` imports anymore, so I replaced that as well.

### DIFF
--- a/.github/workflows/CI_build.yml
+++ b/.github/workflows/CI_build.yml
@@ -93,7 +93,7 @@ jobs:
       - name: Show pip list
         run: |
           pip list
-      - name: Run test with coverage
+      - name: Run test with tensorflow version 2.4
         run: pytest
       - name: Install Tensorflow version 2.5
         run: |
@@ -101,5 +101,5 @@ jobs:
       - name: Show pip list
         run: |
           pip list
-      - name: Run test with coverage
+      - name: Run test with tensorflow version 2.5
         run: pytest
--- a/.github/workflows/CI_build.yml
+++ b/.github/workflows/CI_build.yml
@@ -83,13 +83,21 @@ jobs:
         run: |
           which python
           python --version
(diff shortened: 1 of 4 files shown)


.github/workflows/CI_build.yml
```diff
@@ -69,3 +69,37 @@ jobs:
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install "tensorflow>=2.4,<2.5"
+      - name: Install other dependencies
+        run: |
+          pip install -e .[dev]
+      - name: Show pip list
+        run: |
+          pip list
+      - name: Run test with coverage
```
svenvanderburg, line 96:
```suggestion
      - name: Run test with tensorflow version 2.4
```

.github/workflows/CI_build.yml
```diff
@@ -69,3 +69,37 @@ jobs:
...
+        run: pytest
+      - name: Install Tensorflow version 2.5
+        run: |
+          pip install --upgrade "tensorflow>=2.5,<2.6"
+      - name: Show pip list
+        run: |
+          pip list
+      - name: Run test with coverage
```
svenvanderburg, line 104:
```suggestion
      - name: Run test with tensorflow version 2.5
```

svenvanderburg (approved):
Nice! 
I left 2 very minor improvements.

florian-huber:
Thanks a lot for reviewing, Sven!

PR #95, opened 2021-11-03
Update setup.py

florian-huber:


florian-huber:
Simple change, so only self-reviewing here...

Issue #59, opened 2021-02-10
implement early stopping?


florian-huber:
I ran a model training with early stopping and than got unexpected behavior of `model.save()` since that will now save the final model after training, hence ignoring any `save_best_only=True` setting.

niekdejonge:
@florian-huber 
I believe I have a related issue.
I trained a new model using:

    checkpointer = ModelCheckpoint(filepath=model_file_name, monitor='val_loss', mode="min", verbose=1, save_best_only=True)
    earlystopper_scoring_net = EarlyStopping(monitor='val_loss', mode="min", patience=10, verbose=1)
history = model.model.fit(training_generator, validation_data=validation_generator,
                              epochs=150, verbose=1, callbacks=[earlystopper_scoring_net, checkpointer])

The model that is saved does only store the keras model and not the corresponding spectrum binner. (so it is not using the SiameseModel.save function) 

Possible explanation: I noticed I run model.model.fit() instead of model.fit(), since this returns the history. 

I found a quick hack, by recreating the spectrum_binner and resaving the model:

def save_model_with_spectrum_binner(filename: Union[str, Path], training_spectra):
    with h5py.File(filename, mode='r') as f:
        keras_model = keras.models.load_model(f)
    spectrum_binner = SpectrumBinner(10000, mz_min=10.0, mz_max=1000.0, peak_scaling=0.5,
                                     allowed_missing_percentage=10.0)
    spectrum_binner.fit_transform(training_spectra)
    SiameseModel(spectrum_binner, keras_model=keras_model).save("C:/Users/jonge094/PycharmProjects/PhD_MS2Query/ms2query/data/hot_topics_metabolomics/ms2deepscore_model_with_spectrumbinner.hdf5")

Do you know a way to make sure checkpointer saves the complete model (including spectrum binner) and storing the history?

niekdejonge:
I think I found a solution for the early stopping not saving the best model. Just adding the parameter restore_best_weights=True to the early stopping. (maybe you already suggested this during our meeting) However, I do not really know how to test if this works as expected.
Checkpointer will still have the problem that it does not store the spectrum binning settings, since it uses the original keras save options. So I think the best solution is using early stopping but no checkpointer.
@florian-huber If you agree, you could close this issue

florian-huber:
Outdated.

PR #103, opened 2022-03-28
Create run_ms2deepscore_model_according_to_paper

niekdejonge:
@florian-huber 
The tutorial on MS2Deepscore is a bit outdated. 
It uses different model structure compared to the model in the paper and does not use validation data in the training. 
A student in our group is working with training MS2Deepscore models now, so this might be a good incentive to update this. :) 

The file I added is what I used to train the most recent MS2Deepscore model. Could you check if these settings are indeed correct? In that case we can maybe clean up this code a bit and add explanation and integrate this with the tutorial.

Let me know what you think.


niekdejonge:
In #59 you mention a problem with early stopping, maybe we can implement this right away as well

florian-huber:
As we just discussed, I would leave the early_stopping issue out for now (#59) because that will be a bit of work to implement.

niekdejonge:
When going over the settings I noticed a few settings that I would like to double check:
- Using DataGeneratorAllInchikeys (instead of DataGeneratorAllSpectrums
- Setting allowed_missing_percentage to 100%. Otherwise the spectrum binning of the validation spectra does not work. I believe that having 0 peaks in known bins will always result in an embedding of only zeroes, so this is not a big problem. Also raising an error later, if one spectrum does not fit in is often not wanted behavior.
- I run model.model.fit. In this way the history is stored. However, in the original tutorial model.fit() was run. Is model.model.fit correct?

niekdejonge:
I think I found a solution for the early stopping not saving the best model. Just adding the parameter restore_best_weights=True to the early stopping. (maybe you already suggested this during our meeting) However, I do not really know how to test if this works as expected. 
Checkpointer will still have the problem that it does not store the spectrum binning settings, since it uses the original keras save options. So I think the best solution is using early stopping but no checkpointer.

florian-huber:
> * Using DataGeneratorAllInchikeys (instead of DataGeneratorAllSpectrums
> 

Yes, that's correct!
> * Setting allowed_missing_percentage to 100%. Otherwise the spectrum binning of the validation spectra does not work. I believe that having 0 peaks in known bins will always result in an embedding of only zeroes, so this is not a big problem. Also raising an error later, if one spectrum does not fit in is often not wanted behavior.
> 

Fine as well.
> * I run model.model.fit. In this way the history is stored. However, in the original tutorial model.fit() was run. Is model.model.fit correct?

Regarding the training itself, both will underneath call exactle the same function. The only difference is that the first returns the history, while the second option doesn't (we might change that if possible).

PR #111, opened 2022-11-21
Additional input parameters

djoas:
Able to introduce additional input parameters like "parent_mass" or "precursor_mz".


florian-huber:
After merging main (to fix the recent issues with Tensorflow 2.11, #112), the tests should hopefully pass.

One thing that's missing is adding tests for the additional input option.

florian-huber:
Sorry, tests are not running because it comes from a fork. So I will first merge it into a local branch.

PR #113, opened 2022-11-21
avoid tensorflow 2.11

florian-huber:


florian-huber:
Will use #114 instead.

PR #119, opened 2022-12-22
remove_tf_progress_bar

niekdejonge:
@florian-huber With the update to tensorflow 2.9 the default settings for a progress bar were changed, resulting in printing a progress bar like:
1/1 [==============================] - 0s 223ms/step 
everytime calculate vectors is run. 
This makes all other logging unreadable, especially in MS2Query, since this is called once for each query spectrum. 
This small change should fix this. 

Can I merge this and make a new release?


niekdejonge:
More information can be found here: https://stackoverflow.com/questions/73244216/how-could-hide-the-predict-progress-bar

florian-huber (approved):
Fine for me!
As soon as tests pass, this can be merged.

Issue #118, opened 2022-12-05
multi_inputs is not automatically detected


niekdejonge:
I am running the new multi_inputs pipeline, when running MS2Deepscore, you require to add multi_inputs=True. 

The error you get when forgetting to set this to True is confusing:
/ms2deepscore/MS2DeepScore.py", line 76, in _create_input_vector
    X = np.zeros((1, self.input_vector_dim))
TypeError: 'tuple' object cannot be interpreted as an integer

Two possible solutions:
1. Add an assert statement in MS2Deepscore init, that checks if the input layer is not a tuple. And raises an error suggesting to use multi_inputs =True. 
2. Automatically detect if a model is multi_input. Maybe by storing that it is multi_input as part of the model. Or by detecting the shape of the base model and automatically setting it to multi_input, if the shape is a tuple. 

@florian-huber @djoas Thanks for implementing this :) Let me know if the issue I try to address is unclear!

florian-huber:
Sounds like good suggestions!

2. Would be more comfortable to use, but is probably not backward compatible.
1. Seems easy to implement.

PR #121, opened 2023-01-16
Fix loading multi inputs

florian-huber:
- auto-detect if `multi_inputs` is used
- fix tests.
- linting (`lr` --> `learning_rate` etc.)


florian-huber:
Sorry @djoas, just realized this might be double with changes in the branch https://github.com/matchms/ms2deepscore/tree/additional_input_parameters/ms2deepscore

Anyway, this should also fix #118.

PR #124, opened 2023-03-22
Train models with Metadata

niekdejonge:
@djoas and @florian-huber 

I had a look at making the metadata annotation a bit more robust. What I added was a MetadataFeatureGenerator class.
This contains simple conversions of metadata to input feature, including normalization. This has as benefit that the conversion on the metadata will also be stored in the spectrum binner, so that when the user runs the model later there is no risk of normalizing in a different way for instance. 

@djoas if you are busy with finishing up your thesis and don't have the time. No need to have a look, but since you worked on this it would be great if you notice anything that might give unexpected behavior.


niekdejonge:
@florian-huber Could you have a look at the MetadataFeatureGenerator class I generated. 

Do you think this is a good approach? I never used parent classes before, so wanted to make sure that this would make sense.

In this way we pass a list with Classes of type MetadataFeatureGenerator. In spectrum binning these classes can be used to generate the appropriate metadata based feature. Using the parent class has as benefit that the input stays the same and that they will all contain generate_feature method.

florian-huber:
Nice! I will do a full code review tomorrow.

niekdejonge:
> Nice! I will do a full code review tomorrow.

@florian-huber Great! I noticed that the tensorflow version check was currently failing. Do you know what is causing that? It is unrelated to this pull request right?

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -0,0 +1,28 @@
...
+    def __init__(self, metadata: Metadata):
+        self.metadata = metadata
+
+    def generate_features(self) -> float:
+        """This method should be implemented by child classes to generate a input feature for the model"""
+        raise NotImplementedError
+
+
+class PrecursorMZFeatureGenerator(MetadataFeatureGenerator):
+    def generate_features(self) -> float:
```
florian-huber, line 15:
```suggestion
    """Class for generating feature from precursor-m/z (here: simply m/z divided by 1000).
    """
    def generate_features(self) -> float:
```

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -0,0 +1,28 @@
...
+        precursor_mz = self.metadata.get("precursor_mz")
+        assert precursor_mz is not None, "No precursor mz was found, preprocess your spectra first using matchms"
+        return precursor_mz/1000
+
+
+class IonizationModeFeatureGenerator(MetadataFeatureGenerator):
+    def generate_features(self) -> float:
```
florian-huber, line 22:
```suggestion
    """Class for generating feature from ionization mode. Here simply 0: positive mode and 1: negative mode.
    """
    def generate_features(self) -> float:
```

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -0,0 +1,28 @@
+from matchms import Metadata
+from typing import Tuple
+
+
+class MetadataFeatureGenerator:
+    def __init__(self, metadata: Metadata):
```
florian-huber, line 6:
```suggestion
    """Base class to define metadata-to-feature conversion rules.
    """
    def __init__(self, metadata: Metadata):
```

ms2deepscore/SpectrumBinner.py
```diff
@@ -69,8 +71,14 @@ def from_json(cls, json_str: str):
                               binner_dict["mz_max"], binner_dict["mz_min"],
                               binner_dict["peak_scaling"],
                               binner_dict["allowed_missing_percentage"])
+        # loads in all the classes in MetadataFeatureGenerator.py
         if "additional_metadata" in binner_dict:
-            spectrum_binner.additional_metadata = binner_dict["additional_metadata"]
+            possible_metadata_classes = import_module("ms2deepscore.MetadataFeatureGenerator")
```
florian-huber, line 76: This makes me think.
As it is here users can only use classes defined in ms2deepscore. But one of the main advantages of using a base class `MetadataFeatureGenerator` is that everyone can build their own metadata-to-feature conversions.
Passing custom objects won't work either, because that will be difficult to convert to json and save.
Another problem with this approach is: If people store their model (which includes a binner), but we later decide to change some rules, e.g. m/z divided by 5000 instead of 1000, then loading a model will not fail nor complain but won't work as expected.

One alternative I can think of is to work with more generic conversion mechanisms, say: feature-to-onehot, feature-to-binary, feature-scaler and store all needed factors and choices in a dictionary internally. We could build a class for each such a conversion. The feature-to-binary would need to contain attributes such as `metadata_field`, `entries_becoming_zero`, `entries_becoming_one`...

florian-huber: e.g.
```python
class StandardScaler:
    def __init__(self, metadata_field: str, mean: float, std: float = None):
        self.metadata_field = metadata_field
        self.mean = mean
        self.std = std

    def generate_features(self, metadata):
        feature = metadata.get(self.metadata_field, None)
        assert self.metadata_field is not None, f"Metadata entry for {self.metadata_field} is missing."
        if self.std:
             return (feature - self.mean) / std
        return feature - self.meann
```

florian-huber: We should then add a simple `to_json` method to the respective base class that will export what is needed to reconstruct. I guess that would only be the attributes plus the class name.

florian-huber:
@niekdejonge I made some suggestions for maybe changing the design, let me know if we should have a brief meeting to discuss.

The tensorflow thing should indeed be unrelated. Tensorflow is a constantly moving target.... better to look at that separately.

ms2deepscore/SpectrumBinner.py
niekdejonge: I think that makes a lot of sense. In the end this would not need that many different classes and it makes it robust and easy for users to try other things without the need to clone MS2Deepscore and adding classes to the code. I will have a look into this.

notebooks/train_with_additional_metadata/train_ms2deepscore/old_run_scripts/calculate_tanimoto_scores.py
```diff
@@ -0,0 +1,15 @@
+import os
```
niekdejonge, line 1: Below there are many run functions that were used to create models. I will remove these, since they are not necessary to share on github.

ms2deepscore/data_generators.py
```diff
@@ -476,6 +474,110 @@ def __len__(self):
...
+
+    def __len__(self):
+        """Denotes the number of batches per epoch
+        NB1: self.reference_scores_df only contains 'selected' inchikeys, see `self._data_selection`.
+        NB2: We don't see all data every epoch, because the last half-empty batch is omitted.
+        This is expected behavior, with the shuffling this is OK.
+        """
+        return int(self.settings["num_turns"]) * int(np.floor(len(self.reference_scores_df) / self.settings["batch_size"]))
+
+    def _select_positive_and_negative_mode(self):
```
niekdejonge, line 575: I am not sure if this was checked to have the correct behaviour yet. I believe there are no tests yet, so I don't think this is ready to be merged yet.

ms2deepscore/plotting.py
```diff
@@ -44,15 +44,18 @@ def plot_histograms(histograms, y_score_bins, bin_content=None,
...
+    plt.figure(figsize=(10, 10))
 
+    # Loop over each bin.
     for i in range(0, len(histograms)):
         data = histograms[len(histograms)-i-1][0]
-        data = data/max(data)
-        plt.fill_between(histograms[0][1][:100], -shift*i, [(-shift*i + x) for x in data], color=cmap1(i/10), alpha=alpha)
+        # Normalize the data to have the same area under the curve
+        data = data/sum(data)*len(data)/4
+        plt.fill_between(histograms[0][1][:len(data)], -shift*i, [(-shift*i + x) for x in data], color=cmap1(i/10), alpha=alpha)
```
niekdejonge, line 54: Here the plots are changed to not plot a graph normalized to the highest point, but instead to be normalized to the area under the curve.

ms2deepscore/train_new_model/train_ms2deepscore.py
```diff
@@ -3,44 +3,65 @@
...
+                      epochs=150,
+                      base_dims=(500, 500),
+                      embedding_dim=200
+                      ):
+    assert not os.path.isfile(output_model_file_name), "The MS2Deepscore output model file name already exists"
 
     same_prob_bins = list(zip(np.linspace(0, 0.9, 10), np.linspace(0.1, 1, 10)))
 
-    training_generator = DataGeneratorAllInchikeys(
+    training_generator = DataGeneratorAllSpectrums(
```
florian-huber, line 54: I would still go for `DataGeneratorAllInchikeys`. I looked at the code again and there is indeed something confusing about it. In future PRs I would probably skip the parameter `selected_inchikeys` which only makes things more complicated. Still, the "AllSpectrums" will go through al spectrums which not only makes training longer, it will also shift the bias towards very frequently occurring compounds.

ms2deepscore/train_new_model/train_ms2deepscore.py
```diff
@@ -3,44 +3,65 @@
...
-        selected_inchikeys=list({s.get("inchikey")[:14] for s in training_spectra}),
+        selected_inchikeys=list({s.get("inchikey")[:14] for s in binned_spectrums_training}),
         reference_scores_df=tanimoto_df,
         spectrum_binner=spectrum_binner,
         same_prob_bins=same_prob_bins,
         num_turns=2,
         augment_noise_max=10,
         augment_noise_intensity=0.01)
 
-    validation_generator = DataGeneratorAllInchikeys(
```
florian-huber, line 43: Same as above, I would stick to "AllInchikeys".

notebooks/train_with_additional_metadata/benchmarking/comparison_rmse_plot.py
```diff
@@ -0,0 +1,112 @@
+import os
```
niekdejonge, line 1: Everything in notebooks, does not have to be added in this PR. Still useful locally, but I will remove it from this PR

ms2deepscore/data_generators.py
```diff
@@ -476,6 +474,110 @@ def __len__(self):
         This is expected behavior, with the shuffling this is OK.
         """
         return int(self.settings["num_turns"]) * int(np.floor(len(self.reference_scores_df) / self.settings["batch_size"]))
+    def _spectrum_pair_generator(self, batch_index: int) -> Iterator[SpectrumPair]:
```
niekdejonge, line 477: The _spectrum_pair_generator was not changed/added, this is just Github being confused, since the next class has very similar code

niekdejonge:
@florian-huber 
I went through the pull request and removed all changes that were still in a draft or should only be stored locally. 
The changes remaining in this PR are:
-	In SiameseModel, the attributes are (e.g. self.base) not passed as an argument but instead used by the class. 
-	Improved plotting functionality. Some additional plotting options were added and plots previously created in notebooks are now functions. 
-	The ease of training MS2Deepscore models is improved, including standard settings and splitting validation and training data.

florian-huber:
Thanks @niekdejonge for the update. I went through the code as well and will add a few minor things with #141

florian-huber:
I will merge this since tests are now passing (except for SonarCloud due to missing code coverage of the new workflows, this should be handled elsewhere --> #143 )

niekdejonge:
Great! Thanks for fixing the failing tests!

PR #126, opened 2023-03-24
Generic conversion classes

niekdejonge:
Applied the suggestions of Florian


ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+    @classmethod
+    def load_from_dict(cls, json_dict: dict):
+        """Create StandardScaler instance from json.
+        """
+        return cls(json_dict["metadata_field"],
+                   json_dict["mean"],
+                   json_dict["standard_deviation"],)
+
+
+class OneHotEncode(MetadataFeatureGenerator):
```
florian-huber, line 51:
```suggestion
class OneHotEncoder(MetadataFeatureGenerator):
```

florian-huber: Fits better to the naming of StandardScaler ;)

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+    def generate_features(self, metadata: Metadata):
+        feature = metadata.get(self.metadata_field, None)
+        assert self.metadata_field is not None, f"Metadata entry for {self.metadata_field} is missing."
+        if feature == self.entries_becoming_one:
+            return 1
+        return 0
+
+    @classmethod
+    def load_from_dict(cls, json_dict: dict):
+        """Create StandardScaler instance from json.
```
florian-huber, line 66:
```suggestion
        """Create OneHotEncoder instance from json.
```

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+        if feature == self.entries_becoming_one:
             return 1
-        assert False, "Ionization mode should be 'positive' or 'negative'"
+        if feature == self.entries_becoming_zero:
+            return 0
+        assert False, f"Feature should be {self.entries_becoming_one} or {self.entries_becoming_zero}, not {feature}"
+
+    @classmethod
+    def load_from_dict(cls, json_dict: dict):
+        """Create StandardScaler instance from json.
```
florian-huber, line 91:
```suggestion
        """Create FeatureToBinary instance from json.
```

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+        """
+        return cls(json_dict["metadata_field"],
+                   json_dict["entries_becoming_one"],
+                   json_dict["entries_becoming_zero"],)
+
+
+def load_from_json(json_dict):
```
florian-huber, line 98: Confusing because it is named "from_json", asks for a "dict" but actually takes a list (if I got it right?).

florian-huber, line 98: Isn't it a `json_list` ?

florian-huber: (or other name... but I got confused by the `_dict` when later reading the for-loop)

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+    """Loads any of the objects from load_from_json"""
```
florian-huber, line 99:
```suggestion
    """Create all objects described in load_from_json"""
```

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+    # I think we can directly implement this in the base class.
+    possible_metadata_classes = import_module(__name__)
+    metadata_feature_generator_list = []
+    for metadata_feature_json in json_dict:
+        metadata_feature = json.loads(metadata_feature_json)
+        class_name, settings = metadata_feature
+        # loads in all the classes in MetadataFeatureGenerator.py
+        metadata_class = getattr(possible_metadata_classes, class_name)
+        assert issubclass(metadata_class, MetadataFeatureGenerator)
```
florian-huber, line 108:
```suggestion
        assert issubclass(metadata_class, MetadataFeatureGenerator), "Unknown feature generator class."
```

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+                   json_dict["entries_becoming_one"])
+
+
+class FeatureToBinary(MetadataFeatureGenerator):
+    def __init__(self, metadata_field: str,
+                 entries_becoming_one,
+                 entries_becoming_zero):
+        self.metadata_field = metadata_field
+        self.entries_becoming_one = entries_becoming_one
+        self.entries_becoming_zero = entries_becoming_zero
```
florian-huber, line 78:
```suggestion
                 entries_becoming_one: Union[list, str, int],
                 entries_becoming_zero: Union[list, str, int]):
        self.metadata_field = metadata_field
        if isinstance(entries_becoming_one, list):
            self.entries_becoming_one = entries_becoming_one
        else:
            self.entries_becoming_one = [entries_becoming_one]
        if isinstance(entries_becoming_zero, list):
            self.entries_becoming_zero = entries_becoming_zero
        else:
            self.entries_becoming_zero = [entries_becoming_zero]
```

florian-huber: I think this would make it more generic.

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+
+    def generate_features(self, metadata: Metadata):
+        feature = metadata.get(self.metadata_field, None)
+        assert self.metadata_field is not None, f"Metadata entry for {self.metadata_field} is missing."
+        if feature == self.entries_becoming_one:
             return 1
-        assert False, "Ionization mode should be 'positive' or 'negative'"
+        if feature == self.entries_becoming_zero:
```
florian-huber, line 85:
```suggestion
        if feature in self.entries_becoming_zero:
```

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+                 entries_becoming_one,
+                 entries_becoming_zero):
+        self.metadata_field = metadata_field
+        self.entries_becoming_one = entries_becoming_one
+        self.entries_becoming_zero = entries_becoming_zero
+
+    def generate_features(self, metadata: Metadata):
+        feature = metadata.get(self.metadata_field, None)
+        assert self.metadata_field is not None, f"Metadata entry for {self.metadata_field} is missing."
+        if feature == self.entries_becoming_one:
```
florian-huber, line 83:
```suggestion
        if feature in self.entries_becoming_one:
```

ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -1,34 +1,110 @@
...
+    @classmethod
+    def load_from_dict(cls, json_dict: dict):
+        """Create StandardScaler instance from json.
+        """
+        return cls(json_dict["metadata_field"],
+                   json_dict["entries_becoming_one"])
+
+
+class FeatureToBinary(MetadataFeatureGenerator):
+    def __init__(self, metadata_field: str,
```
florian-huber, line 73:
```suggestion
class CategoricalToBinary(MetadataFeatureGenerator):
    """Converts categorical features (e.g. strings) into a binary 1 or 0 feature.
    """
    def __init__(self, metadata_field: str,
```

florian-huber: I would distinguish between categorical and numerical (or similar). Because it would also make sense to have another version that can create binary features from numbers, say "large molecules" --> 1 for m/z > 500 .

tests/test_data_generators.py
```diff
@@ -183,8 +183,8 @@ def test_DataGeneratorAllSpectrums_additional_inputs():
     spectrums = load_processed_spectrums()
     tanimoto_scores_df = get_reference_scores()
 
-    test_cases = [(PrecursorMZFeatureGenerator, ),  # We are using two test cases, one with only one
-                  (PrecursorMZFeatureGenerator, PrecursorMZFeatureGenerator)]  # Normally you would of course have two different ones (but metadata did not allow it)
+    test_cases = [(StandardScaler("precursor_mz", mean=0, std=1000), ),  # We are using two test cases, one with only one
+                  (StandardScaler("precursor_mz", mean=0, std=1000), StandardScaler("precursor_mz", mean=0, std=100))]  # Normally you would of course have two different ones (but metadata did not allow it)
```
florian-huber, line 187: Above I suggested something like a `NumericalToBinary` generator, which could for instance be used here in the test (m/z >= 500 --> 1, else: 0).

ms2deepscore/MetadataFeatureGenerator.py
niekdejonge: Yes great suggestion!

ms2deepscore/MetadataFeatureGenerator.py
niekdejonge: I agree and will clarify this a bit in a separate commit.

tests/test_data_generators.py
niekdejonge: The current CategoricalToBinary does not exactly do this. It requires to name every possible input value instead of a range. We can later also easily add NumericalToBinary if we want to. For now I will use the CategoricalToBinary for ionization mode as additional test case.

PR #128, opened 2023-04-06
Spectrum binner to data generator

niekdejonge:
Currently the data generator requires the number of bins and additional input. There is a risk that no additional input is applied, while it is used in the spectrum binner, which might result in the user not being aware that this is not used.
To solve this we can pass the spectrum_binner to datagenerator. This also helps since it will not be necessary to specify bin anymore.


niekdejonge:
@florian-huber 
This is an in between PR with the following changes:
- MetaDataFeatureGenerator is added 
- spectrum_binner is passed to data_generator, instead of dim and additional inputs.

florian-huber (approved):

PR #129, opened 2023-04-11
Tensorflow issue

### LINKED ISSUE #123: Saving model non-working
Following MS2DeepScore_tutorial.ipynb,
` model.save(model_file_name)` return ' AttributeError: 'Adam' object has no attribute 'weights' '.
A file is generated but it is unusable, spectrum_binner and additional_input are missing.

- python 3.8.16 (default, Jan 17 2023, 23:13:24)  [GCC 11.2.0]
- matchms 0.18.0 (same issue with 0.16.0)
- ms2deepscore 0.3.1
- tensorflow 2.11.0

niekdejonge:
Fix #123

### DIFF
--- a/ms2deepscore/models/SiameseModel.py
+++ b/ms2deepscore/models/SiameseModel.py
@@ -3,7 +3,6 @@ from typing import Tuple, Union
 import h5py
 from tensorflow import keras
 from tensorflow.keras.layers import BatchNormalization, Dense, Dropout, Input, concatenate  # pylint: disable=import-error
-from tensorflow.python.keras.saving import hdf5_format
 
 from ms2deepscore import SpectrumBinner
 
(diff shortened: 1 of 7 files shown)


ms2deepscore/models/SiameseModel.py
```diff
@@ -112,8 +111,8 @@ def save(self, filename: Union[str, Path]):
             Filename to specify where to store the model.
```
niekdejonge, line 111: This is the real fix of #123

setup.py
```diff
@@ -38,7 +38,8 @@
         "tensorflow;platform_machine!='arm64'",
```
niekdejonge, line 38: @florian-huber Since I added a script for calculating the MS2Deepscore scores rdkit is now also required. I think a good approach could be to separate the running of MS2Deepscore from the training dependencies. What do you think.

niekdejonge: In addition we could also split the tests into two separate test types, to make sure the running pipelines work without the need of the training dependencies.

niekdejonge:
Tensorflow 2.5 does not pass. Why is the tensorflow backwards compatibility required? An alternative solution could be to just force a newer version of tensorflow. Or are the other packages using MS2Deepscore depending on tensorflow?

florian-huber (approved):

florian-huber:
Thanks @niekdejonge !
I merged it. We can follow up on this according to #130.

PR #134, opened 2023-04-21
Clean up data generators

niekdejonge:
PR to clean the code for data generators.
I plan to add new data generators for instance to balance ionisation mode input.
Before doing that I am going through the code to make sure I understand all parts.
During that process I changed things that could be clearer by some refactoring.

@florian-huber let me know what you think.

### DIFF
--- a/ms2deepscore/data_generators.py
+++ b/ms2deepscore/data_generators.py
@@ -411,17 +411,6 @@ class DataGeneratorAllSpectrums(DataGeneratorBase):
 
             np.random.shuffle(self.indexes)
 
-    @staticmethod
-    def _exclude_nans_from_labels(reference_scores_df: pd.DataFrame):
-        """Exclude nans in reference_scores_df, exclude columns and rows if there is any NaN
-        value"""
-        clean_df = reference_scores_df.dropna(axis='rows')  # drop rows with any NaN
-        clean_df = clean_df[clean_df.index]  # drop corresponding columns
-        n_dropped = len(reference_scores_df) - len(clean_df)
-        if n_dropped > 0:
-            print(f"{n_dropped} nans among {len(reference_scores_df)} labels will be excluded.")
-        return clean_df
-
     def _exclude_not_selected_inchikeys(self, reference_scores_df: pd.DataFrame) -> pd.DataFrame:
         """Exclude rows and columns of reference_scores_df for all InChIKeys which are not
         present in the binned_spectrums."""
--- a/ms2deepscore/data_generators.py
+++ b/ms2deepscore/data_generators.py
@@ -69,6 +69,8 @@ class DataGeneratorBase(Sequence):
         use_fixed_set
             Toggles using a fixed dataset, if set to True the same dataset will be generated each
             epoch. Default is False.
+        random_seed
+            Specify random seed for reproducible random number generation.
         """
         self.reference_scores_df = _clean_reference_scores_df(reference_scores_df)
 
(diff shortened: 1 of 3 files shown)


ms2deepscore/data_generators.py
```diff
@@ -226,6 +197,8 @@ def __getitem__(self, batch_index: int):
         if self.settings['use_fixed_set'] and batch_index in self.fixed_set:
             return self.fixed_set[batch_index]
         if self.settings['use_fixed_set'] and batch_index == 0:
+            # todo This sets the global random seed to 42. Wouldn't this cause an issue, since this random seed will
```
niekdejonge, line 200: @florian-huber I am not fully sure but I feel like setting the random seed is a global setting. Therefore any other datagenerator will also use this random seed. 
This would mean that once you use a fixed set datagenerator (which we do for the validation test set). After it has reached this point all other datagenerators will also become fixed, since they will also use seed 42. 

I am not fully sure it this is indeed a global setting and if there is something preventing this, but it certainly felt risky.

niekdejonge: Even if we do not set it in our function there is also a risk that any of the other modules do actually set a random seed. Which would affect how the method functions. 
After googling I found a blog post about this issue including some solutions: 
https://builtin.com/data-science/numpy-random-seed

niekdejonge: hmm, but when calling the np.random function multiple times it will still create different numbers every time. So it might not be an issue in that case. (I have to think about it a bit more :) ).

ms2deepscore/data_generators.py
```diff
@@ -556,3 +518,41 @@ def __init__(self, spectrum_pair, tanimoto_score, dim, _data_augmentation, addit
...
+    value"""
+    clean_df = reference_scores_df.dropna(axis='rows')  # drop rows with any NaN
+    clean_df = clean_df[clean_df.index]  # drop corresponding columns
+    n_dropped = len(reference_scores_df) - len(clean_df)
+    if n_dropped > 0:
+        print(f"{n_dropped} nans among {len(reference_scores_df)} labels will be excluded.")
+    return clean_df
+
+
+def check_duplicated_indexes(reference_scores_df):
```
florian-huber, line 554: Are those functions used individually, outside `clean_reference_scores_df` or the generator classes?

If not, then we could add an underscore to those methods, just to indicate that those are merely helper functions

niekdejonge: No they are not. They were initially inside the classes as static functions. I just did not know that was a good habit for functions outside classes as well. So I will change that.

ms2deepscore/data_generators.py
florian-huber: That is indeed a bit tricky. 
The cleanest way would maybe be to initiate a random number sequence (as in the blog post you mentioned), but since we run a lot of different randomized functions (`choice`, `random`, `randint`) that is maybe not that straightforward.

You can make an issue on this. But since the unit test on this shows that the desired behavior is achieved with the current code I wouldn't change it in this PR.

florian-huber (approved):

ms2deepscore/data_generators.py
florian-huber: Actually, I now added `random_seed` to the settings. Now:
- `fixed_set = True` and `random_seed = ...` --> same fixed order for every generator run and same order for each batch
- only `fixed_set = True`  --> same order for each batch (but different for each new generator initialization and run)

florian-huber:
Thanks a lot @niekdejonge. I added a change regarding the random seed on top. Hope that is in your interest as well.

PR #135, opened 2023-04-21
use dummy data for generator tests

florian-huber:
@niekdejonge this is intended to expand tests on our generators where needed.

In addition the the former tests which use an actual data file it now runs a test using dummy data of the form: 
- fake inchikey "AAAAA..." to "JJJJ....."
- fake tanimoto scores --> simple the distance in the alphabet, scaled to 0 - 1.0
- one peak per spectrum, different position for each inchikey

I hope that this way we can better check if the score bins are used properly. And maybe also test if the data is selected the way we intended.


tests/test_data_generators.py
```diff
@@ -1,22 +1,104 @@
...
     spectrums = load_processed_spectrums()
     tanimoto_scores_df = get_reference_scores()
     ms2ds_binner = SpectrumBinner(100, mz_min=10.0, mz_max=1000.0, peak_scaling=0.5)
     binned_spectrums = ms2ds_binner.fit_transform(spectrums)
     return binned_spectrums, tanimoto_scores_df
 
-
 def test_DataGeneratorAllInchikeys():
-    """Basic first test for DataGeneratorAllInchikeys"""
+    """Basic first test for DataGeneratorAllInchikeys using actual data.
```
niekdejonge, line 60:
```suggestion
    """Test DataGeneratorAllInchikeys using generated data.
```

PR #136, opened 2023-04-25
Generator tests

niekdejonge:
@florian-huber 
I had a closer look to the new tests made and made some changes:
- The dimensions were not set correctly (they were accidentaly) 
- The test if every inchikey was picked only once, did not work. 
- I added a test with dummy data for DataGeneratorAllSpectrums


florian-huber (approved):
Looks good to me, thanks for the changes.

niekdejonge:
Thanks for the quick reply! @florian-huber

PR #146, opened 2023-08-16
Change cherry picking algorithm

niekdejonge:
There is a risk of high fluctuation between the number of pairs in a bin per inchikey.

If you have 5 spectra in a row with no inchikey between 0.9 and 1.0 and the next spectrum has more than a 100, this implementation will result in 100 inchikeys for this inchikey even if the next inchikey also had more matches in this bin.

The behaviour I would prefer is, that it would increase the max bin for all other inchikeys (also the one already calculated). An option could be storing at first 2x max_pairs_per_bin if available and at the end after calculating all tanimoto scores determine how high the max_pair_per_bin should be to reach an average that matches the defined max_pair_per_bin.

This will be a bit more complex in implementation and result in a bit extra overhead and intermediate storage, but I think it is still doable and it reduces the risk of introducing other biases, like oversampling clusters with many similar inchikeys.
It also would make the resulting max_pairs_global be always 0 (unless there is a very extreme distribution).

### DIFF
--- a/ms2deepscore/data_generators.py
+++ b/ms2deepscore/data_generators.py
@@ -5,8 +5,8 @@ from typing import Iterator, List, NamedTuple, Optional
 import numpy as np
 import pandas as pd
 from tensorflow.keras.utils import Sequence  # pylint: disable=import-error
-from ms2deepscore.SpectrumBinner import SpectrumBinner
 from ms2deepscore.spectrum_pair_selection import SelectedCompoundPairs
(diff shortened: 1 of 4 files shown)


florian-huber:
Sorry @niekdejonge, my commits in this PR where supposed to end up in #147 ...
My goal was to make the code easier to read. Feel free to change/suggest edits!

florian-huber (approved):

PR #148, opened 2023-09-13
Update automatic training pipeline

niekdejonge:
- [X] Make pipeline that splits and stores positive and negative mode
- [X] Make pipeline that splits the pos and neg into val, train, test sets
- [x] Fix issue with assert difference == 0 in bias. (the training still crashes)
- [x] Improve docstring
- [x] Add to changelog

Should be implemented in future PR:
- [ ] Fix visualize data. The data visualization needs a Tanimoto df. But this is not calculated anymore. Implement a better way, there was an issue with running the calculate tanimoto matrix, since some of the valid smiles were not able to generate a fingerprint.
- [ ] Add new tests to test the training pipeline. Unit tests have been added, but integration tests for the entire pipeline would still be valuable.


niekdejonge:
@florian-huber There are a few PR's open from a few months ago.
I would like to merge these into a development_training_pipeline branch to prevent creating a long chain of PRs. 
It might help to go over these PR's first because the combined branch will be too massive in the end for easy code review. 
What do you think is a good approach, first going over these smaller PRs or merging everything and later going over the combined PR?

PR #160, opened 2023-11-30
Integrate create_plots with training_wrapper_functions.py

niekdejonge:
The create plots functions were not yet functional, due to a few bugs. 

Functional changes:
- [x] Fix issue with calculate_tanimoto_scores for plotting. The calculations did not work if the df was not symmetric.
- [x] Add and integrate create all plots with training wrapper functions

Internal code changes:
- [x] Added standard file names to SettingsMS2Deepscore
- [x] Make load_negative_train_split (and positive) only return one file at a time, to reduce the risk of assigning the wrong train/test/val variable. 
- [x] Add BenchmarkingResultsFileNames for storing the file names. 

Added tests:
- [x] Make train_ms2ds_wrapper test work. Now it doesn't work, since there are no positive mode spectra and plotting doesn't work. Check that files are created
- [x] Add tests for plotting
- [x] Add test for calculating tanimoto scores for plotting
- [x] Create create_test_spectra.py to store common util functions used by tests, to make updating and reuse easier. 

Folder restructuring
- [x] Created a new folder for all wrapper functions; split plotting function into wrapper functions and tanimoto calculation. 
- [x] Move compute_fingerprints_for_training to spectrum_pair_selection
- [x] Store StoreTrainingData in separate file from train_ms2ds_wrapper 

To do future PR:
- Update Changelog. A more general message that summarizes that we now have a wrapper that does all training, storing all in between changes would be too detailed. 
- Maybe restructure tests in folders as well
- Maybe remove the old data generators
- Store the MS2Deepscore model settings as well (done in #162)


florian-huber:
@niekdejonge I guess this and a few of the following PRs can be closed because they were part of #162 , right?

niekdejonge:
@florian-huber Yes they can be closed!

PR #169, opened 2024-01-17
Integrate pytorch training in train ms2deepscore

niekdejonge:
- Removed tensorflow and all functions depending on it. Some of the old outdated code is still in there and some of the tests are still there (but commented out). Many of these tests will be useless, but there might be some edge cases that they cover that we want to convert to new compatible tests. I did not go through these yet, so I left them in for now, shall we do the further clean up in a later PR?
- Integrated pytorch in train_ms2deepscore
- Updated save and load according to https://pytorch.org/tutorials/beginner/saving_loading_models.html 
- [x] Include tensorize_spectra settings in model. Since these should always be used when running as well. 
- [x] Implement running of the model.


ms2deepscore/models/SiameseSpectralModel.py
```diff
@@ -74,6 +69,24 @@ def forward(self, spectra_tensors_1, spectra_tensors_2, metadata_1, metadata_2):
...
+        ----------
+        filepath: str
+            The file path where the model will be saved.
+        """
+        # Ensure the model is in evaluation mode
+        self.eval()
+        settings_dict = {
+            'model_params': self.model_parameters,
+            'model_state_dict': self.state_dict(),
+            'tensorization_parameters': self.tensorization_parameters.get_dict()
```
florian-huber, line 86: Nice!
That should make things reproducible (or break when needed).

florian-huber (approved):
Very nice @niekdejonge !
We can clean up the remaining tests in a 2nd step.
But this already makes the code base *much* cleaner.

PR #173, opened 2024-01-22
renaming, refactoring, removing old parts

florian-huber:
- remove old parts (binning functions, classes, tests)
- restructure the two classes `SettingsMS2DeepScore` and `GeneratorSettings`
- add `compute_validation_set()` function to later save and re-lead a fixed validation generator.


ms2deepscore/SettingsMS2Deepscore.py
```diff
@@ -132,13 +102,30 @@ class GeneratorSettings:
             Number of pairs for each InChiKey14 during each epoch. Default=1
         shuffle
             Set to True to shuffle IDs every epoch. Default=True
-        ignore_equal_pairs
+        ignore_equal_pairs TODO: check against include_diagonal
```
niekdejonge, line 105: Yes I think this is the same parameter as include_diagonal. But fine to change later.

ms2deepscore/wrapper_functions/training_wrapper_functions.py
```diff
@@ -38,28 +39,28 @@ def train_ms2deepscore_wrapper(spectra_file_path,
...
-                      settings)
+                      model_settings, generator_settings)
 
     # Create performance plots for validation spectra
     ms2deepsore_model_file_name = os.path.join(stored_training_data.trained_models_folder,
-                                               settings.model_directory_name,
-                                               settings.model_file_name)
+                                               model_settings.model_directory_name,
+                                               model_settings.model_file_name)
     calculate_true_values_and_predictions_for_validation_spectra(
```
niekdejonge, line 54: We still have to integrate the use of the datagenerator for validation spectra. It currently works as well, with the implementation I made before: Matching all inchikeys against each other one time, sampling a random spectra for each, repeating a 100 times. But this is also not the exact same over time.

niekdejonge: But this is a future PR, I will add it to the todo list in dev_pytorch

tests/test_data_generators.py
```diff
@@ -177,6 +128,24 @@ def test_DataGeneratorPytorch():
     assert (np.array(counts) > 0.75).sum() > 0.22 * total
 
 
+def test_compute_validation_generator():
+    num_of_unique_inchikeys = 15
```
niekdejonge, line 132: We could add an extra test with saving and loading from pickle. So after creating the val_generator save it by pickling. Create the validation pairs for the original and the pickled one and check that the results are exactly the same.

Issue #89, opened 2021-04-22
Add version information to saved model


florian-huber:
To better handle future compatibility issues it would be nice to track the used version in the saved model. So maybe we can also save some metadata in the model `.hdf5` file.

justinjjvanderhooft:
Good point @florian-huber !!

niekdejonge:
Implemented with #176

PR #176, opened 2024-01-23
Save version

niekdejonge:
Save the version in the model and print a warning if the model is loaded that does not match the current version. 
We can later add more specific warnings (e.g. only add warnings if we know backwards compatibility is broken between versions). But in this way at least all pytorch models will have a version stored, making it backwards compatible if we implement this in more detail later. 
Solves #89


niekdejonge:
Ah yes I created a test model in dev_pytorch, which was not made with version stored, I will update this. (Also why I am happy to add this early on)

PR #183, opened 2024-01-25
Update Changelog

florian-huber:
This is not meant to be merged soon.

I think we should simply collect the main changes here and merge it once we are done with `dev_pytorch`.


niekdejonge:
I checked and could not quickly think of anything that was missing, do you think all the big steps are in there?

PR #186, opened 2024-01-25
fix wrong default

florian-huber:


niekdejonge:
Good catch, I switched halfway through. To make the figure-making and validation set use the same format, so I might have missed some others as well.

PR #187, opened 2024-01-25
add some imports to __init__ (while avoiding circular imports)

florian-huber:


niekdejonge (approved):
Looks good! I cannot think of any other places where we used the +0.0001, so I think you got them all

PR #188, opened 2024-01-25
Add a few tests and minor edits

florian-huber:
- add a few more tests
- minor edits/cleaning
- Add option to have weights for validation loss average.


ms2deepscore/train_new_model/ValidationLossCalculator.py
```diff
@@ -33,7 +36,11 @@ def compute_binned_validation_loss(self,
                                             )
         average_losses = {}
         for loss_type in loss_types:
-            average_losses[loss_type] = np.mean(losses[loss_type])
```
niekdejonge, line 36: @florian-huber This would allow us to do a weighted loss for the validation spectra. But this is not something that is implemented for the training right, or can we weigh the DataGenerator as well? If not, using weights here would result in a loss that is calculated in a different way than the training loss, which would make it very hard to compare the two.

florian-huber: Yes, that's right. I so far used if as a shortcut to test the effect (which wasn't very obvious). I could indeed better add the weighing to the loss functions. Then I would go for a multiplication with a linear function which increases for increasing Tanimotos.

florian-huber: Was now added to the loss functions.

niekdejonge:
Looks good to me!

PR #189, opened 2024-01-25
Integrate settings

niekdejonge:
The settings were not well integrated with the training of the model.
Here the settings are integrated with the training pipeline.

- The model now stores all settings (instead of only model param), this makes sure we store all settings used, so if we later have to check how a model was created this is always stored in the model.
- The SettingsMS2Deepscore, GeneratorSettings, and TensorizationSettings are merged into one


ms2deepscore/models/SiameseSpectralModel.py
```diff
@@ -132,59 +94,43 @@ def output_size(self):
...
-            Default is True in which case the model contains a first dense multi-group peak binning layer.
-        group_size
-            When binning layer is used the group_size determines how many input bins are taken into
-            one dense micro-network.
-        output_per_group
-            This sets the number of next layer bins each group_size group of inputs shares.
-        dropout_rate
-            Dropout rate to be used in the base model.
+        model_settings:
+            SettingsMS2Deepscore object storing all the settings
```
florian-huber, line 105: I think that makes it hard to use for others.

We now put everything together in one huge settings class. This is nice for our current scripts because we can pass the same object to all functions.
But it also somewhat hides the main parameters I think.

I would probably leave it as you suggest for now. But let's add an issue to reconsider the general parameter handling. For instance, it might not be clear to people using it why specific generator settings are also running under model parameters now etc.

ms2deepscore/train_new_model/data_generators.py
```diff
@@ -77,20 +31,8 @@ def __init__(self, spectrums: List[Spectrum],
...
-            Bin width for m/z sampling.
-        intensity_scaling
-            To put more attention on small and medium intensity peaks, peak intensities are
-             scaled by intensity to the power of intensity_scaling.
-        metadata_vectorizer
-            Add the specific MetadataVectorizer object for your data if the model should contain specific
-            metadata entries as input. Default is set to None which means this will be ignored.
-        settings
-            The available settings can be found in GeneratorSettings
+        model_settings
```
florian-huber, line 34: Maybe rename to ms2deepscore_settings or only settings?
I find it confusing that we take the generator settings out of an object model_settings.

niekdejonge: Would training_settings work?

ms2deepscore/models/SiameseSpectralModel.py
niekdejonge: The last changes I made were merging the GeneratorSettings with the Model_settings, so it is quite easy to rollback now. I would be fine with that as well.

niekdejonge: But I am not sure how confused the user would be. I think renaming the SettingsMS2Deepscore and model_settings to TrainingSettings would help a lot (like you suggested below). In that case, it is a nice overview of all the settings used. Having it split with the GeneratorSettings also still works and might make it a bit easier to interpret from just the parameter name what is meant, since it is already clear that this is about data generation.

PR #191, opened 2024-01-29
Add new version of MonteCarlo-Dropout

### LINKED ISSUE #175: Update `MS2DeepScoreMonteCarlo` to pytorch

florian-huber:
This will close #175 .

- Instead of "score" and "uncertainty" the method now returns "score" together with "lower_bound" and "upper_bound" since this is more informative!


niekdejonge (approved):
I like the high and low as well. Uncertainty is hard to interpret or integrate. Low and high is right now missing from the docstring, might be good to add, but not the highest priority

PR #192, opened 2024-01-30
add intensity sum functions

florian-huber:
- added intensity sum functions (just to document them here in case we want to use them later)
- update defaults
- added fingerprint_type and nbits to respective functions where missing.

### DIFF
--- a/ms2deepscore/SettingsMS2Deepscore.py
+++ b/ms2deepscore/SettingsMS2Deepscore.py
@@ -135,11 +135,11 @@ class SettingsMS2Deepscore:
         self.fingerprint_nbits: int = 2048
 
         # Data augmentation
-        self.augment_removal_max = 0.3
+        self.augment_removal_max = 0.2
         self.augment_removal_intensity = 0.2
-        self.augment_intensity = 0.4
+        self.augment_intensity = 0.2
         self.augment_noise_max = 10
-        self.augment_noise_intensity = 0.01
+        self.augment_noise_intensity = 0.02
 
         if settings:
             for key, value in settings.items():
--- a/ms2deepscore/SettingsMS2Deepscore.py
+++ b/ms2deepscore/SettingsMS2Deepscore.py
@@ -100,8 +100,8 @@ class SettingsMS2Deepscore:
         # training settings
         self.dropout_rate = 0.2
         self.learning_rate = 0.00025
-        self.epochs = 150
-        self.patience = 15
+        self.epochs = 250
+        self.patience = 30
         self.loss_function = "mse"
         self.weighting_factor = 0
 
--- a/ms2deepscore/SettingsMS2Deepscore.py
+++ b/ms2deepscore/SettingsMS2Deepscore.py
@@ -88,7 +88,7 @@ class SettingsMS2Deepscore:
         """
     def __init__(self, **settings):
         # model structure
-        self.base_dims = (1000, 1000)
+        self.base_dims = (2000, 2000, 2000)
         self.embedding_dim = 400
         self.ionisation_mode = "positive"
 
@@ -101,7 +101,7 @@ class SettingsMS2Deepscore:
         self.dropout_rate = 0.2
         self.learning_rate = 0.00025
         self.epochs = 150
-        self.patience = 10
+        self.patience = 15
         self.loss_function = "mse"
 
         # Folder names for storing
--- a/ms2deepscore/SettingsMS2Deepscore.py
+++ b/ms2deepscore/SettingsMS2Deepscore.py
@@ -103,6 +103,7 @@ class SettingsMS2Deepscore:
         self.epochs = 150
         self.patience = 15
         self.loss_function = "mse"
+        self.weighting_factor = 0
 
         # Folder names for storing
         self.model_file_name = "ms2deepscore_model.pt"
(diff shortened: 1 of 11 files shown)


ms2deepscore/SettingsMS2Deepscore.py
```diff
@@ -101,8 +101,9 @@ def __init__(self, **settings):
         self.dropout_rate = 0.2
         self.learning_rate = 0.00025
         self.epochs = 150
-        self.patience = 10
+        self.patience = 15
```
niekdejonge, line 104: I normally use an even higher patience, but this doesn't have to be exactly the same for both of us. The rest looks like good defaults.

niekdejonge: We could also turn of dropout by default?

ms2deepscore/benchmarking/calculate_scores_for_validation.py
```diff
@@ -48,13 +48,19 @@ def calculate_true_values_and_predictions_for_validation_spectra(positive_valida
 
 
 def calculate_tanimoto_scores_unique_inchikey(list_of_spectra_1: List[Spectrum],
-                                              list_of_spectra_2: List[Spectrum]):
+                                              list_of_spectra_2: List[Spectrum],
+                                              fingerprint_type="daylight",
+                                              nbits=2048
+                                              ):
     """Returns a dataframe with the tanimoto scores between each unique inchikey in list of spectra"""
```
niekdejonge, line 55: Great, I must have missed that during integration

ms2deepscore/train_new_model/data_generators.py
```diff
@@ -156,7 +156,7 @@ def _data_augmentation_spectrum(self, spectrum_tensor):
             # TODO: Factor out function with documentation + example?
             
             indices_select = torch.where((spectrum_tensor > 0)
-                                         & (spectrum_tensor < self.model_settings.augment_removal_max))[0]
+                                         & (spectrum_tensor < self.model_settings.augment_removal_intensity))[0]
```
niekdejonge, line 159: What is the difference here? Because the augment_removal_max is also still in the settings

ms2deepscore/train_new_model/train_ms2deepscore.py
```diff
@@ -42,6 +42,7 @@ def train_ms2ds_model(
     model = SiameseSpectralModel(settings=settings)
 
     validation_loss_calculator = ValidationLossCalculator(validation_spectra,
+                                                          settings=settings,
                                                           score_bins=settings.same_prob_bins,
```
niekdejonge, line 46: We could remove the settings.same_prob_bins and random_seed here as well and use settings in the init.

ms2deepscore/utils.py
```diff
@@ -37,3 +39,32 @@ def load_spectra_as_list(file_name) -> List[Spectrum]:
     if isinstance(spectra, Generator):
         return list(spectra)
     return spectra
+
+
+@numba.jit(nopython=True)
+def scaled_intensity_sum(mz_values, intensities, min_mz=0, max_mz=1000, scaling=2):
+    """Compute a scaled intensity sum for all peaks of a spectrum.
+
+    Each peak will be scaled by the power of `scaling`.
```
niekdejonge, line 48: Seems to do what it says. Did this improve performance for you? The rational behind it is not directly apparent to me (but that is fine as long as you know)

tests/test_siamese_spectra_model.py
```diff
@@ -138,7 +138,8 @@ def test_model_training(simple_training_spectra):
 
     validation_loss_calculator = ValidationLossCalculator(
         simple_training_spectra,
-        [(0, 1.00001),]) # Just calculating the loss in one bin (since we have just two inchikeys)
+        settings=settings,
+        score_bins=[(0, 1.00001),]) # Just calculating the loss in one bin (since we have just two inchikeys)
```
niekdejonge, line 142: Can be switched to 0, 1 as well (missed in previous PR) Or just remove at all and make part of settings

niekdejonge (approved):
Looks good! I had some small additions to make some of the parts complete, but I will quickly implement them myself and merge.

ms2deepscore/train_new_model/data_generators.py
niekdejonge: @florian-huber

ms2deepscore/utils.py
niekdejonge: @florian-huber

niekdejonge: Ah I misunderstood what the purpose was, but now get it again. this was the function to remove spectra that did not have enough high intensity peaks, for spectra with many inchikeys for instance. Maybe quickly add that to the docstring as well

ms2deepscore/train_new_model/ValidationLossCalculator.py
```diff
@@ -8,21 +8,28 @@
...
     def __init__(self,
                  val_spectrums,
-                 score_bins=np.array([(x / 10, x / 10 + 0.1) for x in range(0, 10)]),
-                 weighting_factor=0,
+                 settings: SettingsMS2Deepscore,
+                 score_bins=None,
                  random_seed=42):
         self.val_spectrums = select_one_spectrum_per_inchikey(val_spectrums, random_seed)
-        tanimoto_scores = get_tanimoto_score_between_spectra(self.val_spectrums, self.val_spectrums)
+        if score_bins is None:  # Use same as for training
```
niekdejonge, line 21: Don't we always want to use the same score bins here?

ms2deepscore/train_new_model/data_generators.py
florian-huber: Only peaks with intensities < `augment_removal_intensity` are selected for potential removal.
Out of those, a fraction of 0 to `removal_max` can be removed.

niekdejonge: Ah so this was a bug before?

florian-huber:
Thanks for reviewing and additional changes @niekdejonge !

PR #194, opened 2024-01-31
Small fixes

niekdejonge:
Almost no functional change except for fixing some bugs with the metadataGenerator integration.
- For the model file name generation, the way we store the metadataGenerator slightly changed, was adjusted accordingly in the pipeline.
- I enabled metadata harmonization again when loading spectra. Otherwise, the precursor mz is stored as string which breaks the metadata generator. Metadata harmonization was disabled before, because it would use pepmass to overwrite previously fixed precursor mz. But I believe this was fixed with a newer version of matchms.

### DIFF
--- a/ms2deepscore/MetadataFeatureGenerator.py
+++ b/ms2deepscore/MetadataFeatureGenerator.py
@@ -82,7 +82,7 @@ class StandardScaler(MetadataFeatureGenerator):
     def generate_features(self, metadata: Metadata):
         feature = metadata.get(self.metadata_field, None)
         assert self.metadata_field is not None, f"Metadata entry for {self.metadata_field} is missing."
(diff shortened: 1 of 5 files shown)


ms2deepscore/MetadataFeatureGenerator.py
```diff
@@ -81,8 +81,10 @@ def __init__(self, metadata_field: str, mean: float, std: float = None):
 
     def generate_features(self, metadata: Metadata):
         feature = metadata.get(self.metadata_field, None)
-        assert self.metadata_field is not None, f"Metadata entry for {self.metadata_field} is missing."
```
florian-huber, line 84: I changed this into raise-constructions.

florian-huber (approved):

PR #195, opened 2024-01-31
Fix suggestions

niekdejonge:


ms2deepscore/train_new_model/ValidationLossCalculator.py
```diff
@@ -14,14 +14,9 @@
...
                  val_spectrums,
-                 settings: SettingsMS2Deepscore,
-                 score_bins=None,
-                 random_seed=42):
-        self.val_spectrums = select_one_spectrum_per_inchikey(val_spectrums, random_seed)
-        if score_bins is None:  # Use same as for training
-            self.score_bins = settings.same_prob_bins
-        else:
-            self.score_bins = score_bins
+                 settings: SettingsMS2Deepscore):
```
niekdejonge, line 17: I think we would always want to use the same_prob_bins that were used during training here. So to reduce the risk of introducing errors, this might be best to remove as an option. (if different prob_bins, val loss cannot be compared to train loss)

florian-huber: Thats fine with me , too!

ms2deepscore/SettingsMS2Deepscore.py
```diff
@@ -98,10 +98,10 @@ def __init__(self, **settings):
         self.train_binning_layer_output_per_group: int = 2
 
         # training settings
-        self.dropout_rate = 0.2
+        self.dropout_rate = 0.0
         self.learning_rate = 0.00025
-        self.epochs = 150
-        self.patience = 15
+        self.epochs = 250
+        self.patience = 30
```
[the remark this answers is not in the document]
florian-huber: I left dropout. So far my results are a bit inconclusive but showed a several training runs seemed to work slightly better with dropout.

florian-huber (approved):

PR #198, opened 2024-02-01
Cleanup wrappers

florian-huber:


niekdejonge (approved):
Looks good to me! For now we just leave in both methods. But in the end we can probably best keep one of the two to keep the codebase simpler.

Issue #199, opened 2024-02-01
Mac 1 chip results in wrong results


niekdejonge:
@kevinmildau notified me that there was a potential issue with the M1 chip. The predictions are notably different when run on an macos with M1 chip compared to windows, without any warnings or errors. 

The M1 chip is made compatible by installing tensorflow-macos (instead of tensorflow)
This made it run on macos with M1 chip, unfortunately, automatic testing of the M1 chip is not available yet on github actions. So we would not notice if this is the case.

kevinmildau:
Using a conda environment with python 3.8 and pytest 8.0.0 on macbook Pro M1 Pro from 2021

![image](https://github.com/matchms/ms2deepscore/assets/32129512/e614261e-c225-40f8-933e-5e02abb4cec2)

niekdejonge:
Proposed short term solution:
Remove tensorflow-macos in setup.py. This will make it impossible to install on macos systems with the M1 chip.

Long term solution:
We switched to pytorch, which is hopefully better compatible with the M1 chip.

florian-huber:
Fixed with #200 (and even more so with version 2.0).

PR #200, opened 2024-02-01
Remove M1 chip compatibility

niekdejonge:
As described in #199 currently MS2Deepscore runs on macos with the M1 chip, but it makes drastically different predictions than on windows.
Not sure what the issue is, but for now we should just block installing. We will probably not solve this for the tensorflow version, but hopefully this will just not be an issue with pytorch.


niekdejonge:
@florian-huber Do you agree with this (temporary) solution?

florian-huber (approved):

PR #206, opened 2024-02-27
Integrate train in class

niekdejonge:
Small changes to improve clarity of embeddingsEvaluator

- Added train_evaluator as method to EmbeddingEvaluatorModel
- Added test for training EmbeddingEvaluator

PR #207, opened 2024-02-28
Split SettingsMS2Deepscore and SettingsEvaluationModel

niekdejonge:
Continues in changes made in #206, I made a new PR to keep a clear overview. 

Here I split the SettingsMS2Deepscore and SettingsEvaluationModel. Let me know what you think.

### DIFF
--- a/ms2deepscore/SettingsMS2Deepscore.py
+++ b/ms2deepscore/SettingsMS2Deepscore.py
@@ -194,10 +194,10 @@ class SettingsEmbeddingEvaluator:
 
         # Training settings
         # todo set actually good settings
-        self.mini_batch_size = 32
-        self.batches_per_iteration = 100
-        self.learning_rate = 0.001
-        self.num_epochs = 200
+        self.mini_batch_size = 100
+        self.batches_per_iteration = 1000
+        self.learning_rate = 0.0001
+        self.num_epochs = 5
 
         if settings:
             for key, value in settings.items():
--- a/ms2deepscore/SettingsMS2Deepscore.py
+++ b/ms2deepscore/SettingsMS2Deepscore.py
@@ -190,6 +190,13 @@ class SettingsEmbeddingEvaluator:
         self.evaluator_kernel_size = 20
         self.random_seed: Optional[int] = None
 
+        # Training settings
+        # todo set actually good settings
+        self.mini_batch_size = 32
+        self.batches_per_iteration = 100
+        self.learning_rate = 0.001
+        self.num_epochs = 200
+
         if settings:
             for key, value in settings.items():
                 if hasattr(self, key):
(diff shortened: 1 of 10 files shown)


ms2deepscore/train_new_model/spectrum_pair_selection.py
```diff
@@ -150,13 +152,14 @@ def compute_fingerprint_dataframe(
         The settings that should be used for selecting the compound pairs wrapper. The settings should be specified as a
         SettingsMS2Deepscore object.
     """
-    if settings.random_seed is not None:
-        np.random.seed(settings.random_seed)
+    # todo why do we need to set a random seed here??
```
niekdejonge, line 155: @florian-huber Why would we need a random seed here? The fingerprint calculations don't have a random component right?

florian-huber: True. Must be a leftover and can be removed!

ms2deepscore/SettingsMS2Deepscore.py
```diff
@@ -183,3 +177,35 @@ def default(self, o):
...
+        self.evaluator_depth = 3
+        self.evaluator_kernel_size = 20
+        self.random_seed: Optional[int] = None
+
+        # Training settings
+        # todo set actually good settings
+        self.mini_batch_size = 32
+        self.batches_per_iteration = 100
+        self.learning_rate = 0.001
+        self.num_epochs = 200
```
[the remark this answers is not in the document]
florian-huber: Still maybe not the final state, but probably closer...

ms2deepscore/models/EmbeddingEvaluatorModel.py
```diff
@@ -68,46 +70,41 @@ def save(self, filepath):
...
 
-    def train_evaluator(
-            self,
-            data_generator: DataGeneratorEmbeddingEvaluation,
-            mini_batch_size: int,
-            batches_per_iteration: int,
-            learning_rate: float,
-            num_epochs: int,
-            val_generator: DataGeneratorEmbeddingEvaluation = None):
+    def train_evaluator(self,
```
florian-huber, line 73:
```suggestion
    def train(self,
```

florian-huber: Since it is now part of the model class we don't have to specify what this is training.

niekdejonge: There is a method of the nn.module which is also called train, which we call in train_evaluator. Which I think would result in us breaking something, but I am not fully sure.

florian-huber: Oh sorry, that's of course correct. Than nevermind....

PR #215, opened 2024-04-19
Export for java

niekdejonge:
To make MS2deepscore with MZMine we will need to export the model in a different way


niekdejonge:
This has now been implemented in https://github.com/niekdejonge/ms2deepscore_java_conversion

PR #228, opened 2024-08-08
Make relu a setting

niekdejonge:


niekdejonge:
@florian-huber I made the activation function a setting. Instead of defaulting everything to relu.

PR #231, opened 2024-08-29
Fix selected compound pairs bug

niekdejonge:
When having smaller datasets there is a chance that a inchikey does not have available pairs in the datagenerator. This will result in an index out of bound error. This can happen because an inchikey can already occur frequently in the list of inchikeys to be selected by other pairs. Here we implemented a test and skip empty pair lists for inchikeys. 

This happens during the function balanced_selection if the minimum number of pairs per bin is < number of inchikeys. This is the case for the validation set. Nr of inchikeys is 1699, but min pairs per bin (0.8-0.9 tanimoto) is 944. This means that for all the bins only 944 pairs are selected. So each inchikey starts with 0 pairs that are than filled. By chance it can happen that one inchikey is selected 0 times. 

As long as the number of pairs per bin is quite a bit higher than the number of inchikeys this leads to quite a good balancing of the inchikeys. The inchikey balancing will become better if we set max_oversampling rate to a higher number.

### DIFF
--- a/ms2deepscore/train_new_model/data_generators.py
+++ b/ms2deepscore/train_new_model/data_generators.py
@@ -87,7 +87,10 @@ class DataGeneratorPytorch:
         indexes = self.indexes[batch_index * batch_size:(batch_index + 1) * batch_size]
         for index in indexes:
             inchikey1 = self.selected_compound_pairs.idx_to_inchikey[index]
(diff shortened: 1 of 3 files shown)


niekdejonge:
This PR is superseded by #232 The changes made here are made unnecessary (and changed again) by the changes made in #232

PR #232, opened 2024-08-29
Improve inchikey pair selection and data generators

niekdejonge:
Refactoring of the InchikeyPairSelection and datagenerators. 

Changes made:
- Made DataGenerator and SelectedCompoundPairs less intertwined. SelectedCompoundPairs just contains the inchikey pairs and the dataGenerator just adds the spectrum picking and tensorization.
- Fixed the bug that only one spectrum per inchikey was used during training
- Sampling of inchikeys is more balanced, the algorithm is now like this:
  - First tanimoto scores are collected per tanimoto score bin a 100 pairs are selected for each inchikey (if available). The 100 pairs per inchikey is used to keep the algorithm scalable and not run into memory issues, in this way memory scales linearly instead of quadratically. 
  - Pairs are converted from a matrix to just a list of pairs in the form (inchikey1, inchikey2, score). Any inverted duplicates are removed (e.g. (inchikey2, inchikey1, score)). 
  - We count the available number of pairs per bin. The number of pairs in the bin that has the least number of bins is used as pairs per bin, to ensure a good balance over the different bins. 
  - We start selecting pairs per bin from the bin with the lowest number of pairs to the highest number of pairs.
  - We track the frequency of each inchikey in all pairs over all bins. (both the first and the second inchikey in each pair is counted, which was not the case before). 
  - When picking pairs we start with the inchikey with the lowest frequency that is available for that tanimoto bin. We select all available pairs for this inchikey in this bin. There can be mulitple pairs available. (e.g. [(inchikey5, inchikey3, 0.9), (inchikey5, inchikey8, 0.92), (inchikey5, inchikey6, 0.91)]. From these pairs we pick the pair were the second inchikey is the least frequent. 
  - This results in a list of pairs, that have a perfect balance over the different bins, with an almost perfect balance of unique inchikeys. 

Additional changes:
- I have improved the tests, we now test the balance of inchikeys, the balance of scores, and that we actually select all spectra (and not just one per inchikey). 
- Before an epoch would be exactly nr_of_inchikeys long. Even if it did not fit with the batch size. This could result in the final batch being smaller than the batch_size. Now I have changed this to just using the batch size for the final batch size as well, resulting in an epoch that takes a few more pairs than the nr_of_inchikeys.


tests/test_siamese_spectra_model.py
```diff
@@ -125,17 +125,17 @@ def test_model_training(simple_training_spectra):
                                     intensity_scaling=0.5, base_dims=(200, 200),
                                     embedding_dim=100,
                                     train_binning_layer=False,
-                                    same_prob_bins=np.array([(0, 1.0)]),
+                                    same_prob_bins=np.array([(0, 0.5), (0.5, 1.0)]),
```
niekdejonge, line 128: The last tests checking: assert (np.array(history["collection_targets"]) == 1).sum() == 200
did pass before. However, this was as far as I understand just coincidence, since no same_prob_bins were set. So the pair distribution can be random. Since we balance the inchikey picking better now, this was 266. 
I did correctly understand this test right?

tests/test_siamese_spectra_model.py
```diff
@@ -125,17 +125,17 @@ def test_model_training(simple_training_spectra):
...
                                     num_turns=20,
                                     )
-    scp_train, _ = select_compound_pairs_wrapper(simple_training_spectra, settings)
+    scp_train = select_compound_pairs_wrapper(simple_training_spectra, settings)
 
     # Create generators
     train_generator_simple = DataGeneratorPytorch(spectrums=simple_training_spectra, selected_compound_pairs=scp_train,
                                                   settings=settings)
-
+    settings.same_prob_bins = np.array([(0, 1.0)])
```
niekdejonge, line 138: The validation loss calculator does not work for such a small dataset with two bins. So I set it here to 1 bin. 
I did not really check if it makes sense that it doesn't work with two bins, but since I noticed that some changes need to be made anyway, I will have a look at this in a different PR.

tests/test_spectrum_pair_selection.py
niekdejonge, line 1: It was actually renamed to test_inchikey_pair_selection, with additional changes. Not all tests were replaced

ms2deepscore/train_new_model/data_generators.py
```diff
@@ -9,23 +9,27 @@
...
-    have higher significance in the training dataset.
+    This class provides a data generator specifically designed for training a Siamese Pytorch model with a curated set
+    of compound pairs. It takes a SelectedInchikeyPairs and randomly selects, augments and tensorizes spectra for each
+    inchikey pair.
+
+    By using pre-selected compound pairs (in the SelectedInchikeyPairs), this allows more control over the training
+    process. The selection of inchikey pairs does not happen in DataGeneratorPytorch (only spectrum selection), but in
+    inchikey_pair_selection.py. In inchikey_pair_selection inchikey pairs are picked to balance selected pairs equally
+    over different tanimoto score bins to make sure both pairs of similar and dissimilar compounds are sampled.
+    In addition inchikeys are selected to occur equally for each pair.
```
florian-huber, line 28:
```suggestion
    `SelectedInchikeyPairs` (which will take care of picking well-balanced inchikey pairs).
```

florian-huber: I think the more detailed description should be moved to the respective classes/functions.

ms2deepscore/train_new_model/inchikey_pair_selection.py
```diff
@@ -120,44 +70,106 @@ def select_compound_pairs_wrapper(
...
-        fingerprint_nbits)
-
-    fingerprints_df = pd.DataFrame(fingerprints, index=inchikeys14_unique)
-    return fingerprints_df
+    available_pairs_per_bin = convert_selected_pairs_matrix(available_pairs_per_bin_matrix, available_scores_per_bin_matrix, inchikeys14_unique)
+    selected_pairs_per_bin = balanced_selection_of_pairs_per_bin(available_pairs_per_bin, inchikeys14_unique)
+    return SelectedInchikeyPairs([pair for pairs in selected_pairs_per_bin for pair in pairs])
+
+
+def convert_selected_pairs_matrix(selected_pairs_per_bin_matrix, scores_per_bin, inchikeys) -> List[List[Tuple[str, str, float]]]:
```
florian-huber, line 89: Would need a docstring to explain what this function does.

ms2deepscore/train_new_model/inchikey_pair_selection.py
```diff
@@ -120,44 +70,106 @@ def select_compound_pairs_wrapper(
...
+            inchikey_2 = inchikeys[selected_pairs_per_bin_matrix[bin_idx, inchikey_index_1, pair_sample_position[i]]]
+            score = scores_per_bin[bin_idx, inchikey_index_1, pair_sample_position[i]]
+            # Check that the reversed pair (which is the same) is not already in the set
+            if (inchikey_2, inchikey_1, score) not in pairs:
+                pairs.append((inchikey_1, inchikey_2, score))
+        selected_pairs_per_bin.append(pairs)
+    return selected_pairs_per_bin
+
+
+def balanced_selection_of_pairs_per_bin(list_of_pairs_per_bin,
```
florian-huber, line 105: Would need a docstring to explain what this function does.

ms2deepscore/train_new_model/inchikey_pair_selection.py
```diff
@@ -120,44 +70,106 @@ def select_compound_pairs_wrapper(
...
+                                                   key=lambda i: len(list_of_pairs_per_bin[i]))
+    lowest_number_of_pairs = min(len(pairs) for pairs in list_of_pairs_per_bin)
+    selected_pairs_per_bin = []
+    for bin_index in tqdm(sorted_bin_indices_on_amount_of_pairs):
+        selected_pairs_per_bin.append(
+            select_pairs(list_of_pairs_per_bin[bin_index], inchikey_count, lowest_number_of_pairs))
+    return selected_pairs_per_bin
+
+
+def select_pairs(list_of_available_pairs, inchikey_counts, required_number_of_pairs):
```
florian-huber, line 118: Would need a docstring to explain what this function does.

ms2deepscore/train_new_model/inchikey_pair_selection.py
```diff
@@ -120,44 +70,106 @@ def select_compound_pairs_wrapper(
...
+        # Remove the selected pair and a potential reversed pair if available
+        if (idx_1, idx_2, score) in list_of_available_pairs:
+            list_of_available_pairs.remove((idx_1, idx_2, score))
+        if (idx_2, idx_1, score) in list_of_available_pairs:
+            list_of_available_pairs.remove((idx_2, idx_1, score))
+
+        selected_pairs.append((idx_1, idx_2, score))
+        inchikey_counts[idx_1] += 1
+        inchikey_counts[idx_2] += 1
+    return selected_pairs
```
florian-huber, line 140: `inchikey_counts` is given as a parameter and updated in this function --> should also be returned

florian-huber: Ahh. Sorry, I see now that this is only needed with within the for-loop. I would then better make this clearer in the code (I'll add a suggestion).

ms2deepscore/train_new_model/inchikey_pair_selection.py
```diff
@@ -120,44 +70,106 @@ def select_compound_pairs_wrapper(
...
+            # Check that the reversed pair (which is the same) is not already in the set
+            if (inchikey_2, inchikey_1, score) not in pairs:
+                pairs.append((inchikey_1, inchikey_2, score))
+        selected_pairs_per_bin.append(pairs)
+    return selected_pairs_per_bin
+
+
+def balanced_selection_of_pairs_per_bin(list_of_pairs_per_bin,
+                                        unique_inchikeys):
+    inchikey_count = {inchikey: 0 for inchikey in unique_inchikeys}
```
florian-huber, line 107:
```suggestion
```

ms2deepscore/train_new_model/inchikey_pair_selection.py
```diff
@@ -120,44 +70,106 @@ def select_compound_pairs_wrapper(
...
+    sorted_bin_indices_on_amount_of_pairs = sorted(range(len(list_of_pairs_per_bin)),
+                                                   key=lambda i: len(list_of_pairs_per_bin[i]))
+    lowest_number_of_pairs = min(len(pairs) for pairs in list_of_pairs_per_bin)
+    selected_pairs_per_bin = []
+    for bin_index in tqdm(sorted_bin_indices_on_amount_of_pairs):
+        selected_pairs_per_bin.append(
+            select_pairs(list_of_pairs_per_bin[bin_index], inchikey_count, lowest_number_of_pairs))
```
florian-huber, line 114:
```suggestion
            select_pairs(list_of_pairs_per_bin[bin_index], unique_inchikeys, lowest_number_of_pairs))
```

ms2deepscore/train_new_model/inchikey_pair_selection.py
florian-huber, line 118:
```suggestion
def select_pairs(list_of_available_pairs, unique_inchikeys, required_number_of_pairs):
```

ms2deepscore/train_new_model/inchikey_pair_selection.py
```diff
@@ -120,44 +70,106 @@ def select_compound_pairs_wrapper(
...
+    return selected_pairs_per_bin
+
+
+def select_pairs(list_of_available_pairs, inchikey_counts, required_number_of_pairs):
+    selected_pairs = []
+    for _ in range(required_number_of_pairs):
```
florian-huber, line 120:
```suggestion
    selected_pairs = []
    inchikey_count = {inchikey: 0 for inchikey in unique_inchikeys}
    for _ in range(required_number_of_pairs):
```

florian-huber: This is to make clearer that the counter is only need within this function.

ms2deepscore/train_new_model/inchikey_pair_selection.py
niekdejonge: No you are correct! I try to achieve balanced inchikey counts over the different bins. The code does this currently, but it is always confusing and dangerous to update dictionaries within a function without returning. So I fully agree this should be changed.

PR #233, opened 2024-09-25
Cleaning refactoring 232

florian-huber:
Refactoring a few functions for readability (although that is always a bit subjective...) and speed. Plus minor changes.

### DIFF
--- a/ms2deepscore/SettingsMS2Deepscore.py
+++ b/ms2deepscore/SettingsMS2Deepscore.py
@@ -207,34 +207,32 @@ class SettingsMS2Deepscore:
 
 def validate_bin_order(score_bins):
     """Checks that the given bins are of the correct format
-
     The bins should cover everything between 0 and 1.0 and the lowest bin should be below 0
     (since pairs > are selected and we want to include zero)"""
     # check that the correct same_prob_bins are selected
-    bin_borders_below_zero = 0
-    bin_borders_1 = 0
-    not_starting_or_ending_borders = []
-    for score_bin in score_bins:
-        if score_bin[0] > score_bin[1]:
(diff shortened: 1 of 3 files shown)


ms2deepscore/SettingsMS2Deepscore.py
```diff
@@ -206,35 +205,37 @@ def default(self, o):
 
 
 def validate_bin_order(score_bins):
-    """Checks that the given bins are of the correct format
+    """
```
niekdejonge, line 208: Thanks for making this code a bit clearer. This reads logical for me as well.

ms2deepscore/train_new_model/inchikey_pair_selection.py
```diff
@@ -250,83 +251,110 @@ def select_balanced_pairs(available_pairs_for_bin_matrix: np.ndarray,
...
             nr_of_pairs_selected += 1
             progress_bar.update(1)
+
+            # If this inchikey still has valid pairs, push it back into the heap
+            if np.any(pair_frequency[inchikey_with_lowest_count] < max_resampling):
+                heapq.heappush(available_inchikey_indexes,
+                               (inchikey_counts[inchikey_with_lowest_count], inchikey_with_lowest_count))
+
+            # Also check if the second inchikey needs to be added back to the heap
+            if second_inchikey_with_lowest_count != inchikey_with_lowest_count:
```
niekdejonge, line 353: The second inchikey is never popped from the available inchikey indexes, so I believe it should never be added back in.

niekdejonge (changes requested):
Looks good, this is indeed also clearer to me. And I can imagine that the refactoring will result in some speed up. I have made some suggested changes in #234

niekdejonge (approved):

Issue #246, opened 2024-10-22
fingerprint bits default cannot be changed


florian-huber:
The fingerprint size (nbits) can be changed in settings, but this is not correctly passed on in `calculate_true_values_and_predictions_for_validation_spectra` (so it will always use the default of 2048).

PR #250, opened 2024-10-29
Minor code linting

florian-huber:
Since #246 was fixed in #244 in the end, this was only about minor additional linting.


florian-huber:
Only some minor linting.

PR #251, opened 2024-10-30
Use fingerprints for plotting

niekdejonge:
Make the benchmarking plots use fingerprint settings as well, to ensure the same settings are used as during training. (in the automated training pipeline)


florian-huber (approved):
Look good, thanks @niekdejonge !

PR #253, opened 2024-11-07
Update sampling algorithm

niekdejonge:
Added a max inchikey sampling as setting

To make it a bit more consistent I changed average_pairs_per_bin to average_inchikey_sampling_count

### DIFF
--- a/ms2deepscore/SettingsMS2Deepscore.py
+++ b/ms2deepscore/SettingsMS2Deepscore.py
@@ -135,16 +135,15 @@ class SettingsMS2Deepscore:
         self.use_fixed_set = False
 
         # Compound pairs selection settings
-        self.average_pairs_per_bin = 20
-        # todo replace average pairs per bin to. Average nr of sampling per inchikey
-        self.max_inchikey_sampling = 400
+        self.average_inchikey_sampling_count = 400
+        self.max_inchikey_sampling = 440
         self.max_pairs_per_bin = 100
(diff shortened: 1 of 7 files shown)


ms2deepscore/SettingsMS2Deepscore.py
```diff
@@ -135,14 +135,15 @@ def __init__(self, validate_settings=True, **settings):
...
+        self.average_inchikey_sampling_count = 400
+        self.max_inchikey_sampling = 440
         self.max_pairs_per_bin = 100
         self.same_prob_bins = np.array([(0.8, 0.9), (0.7, 0.8), (0.9, 1.0), (0.6, 0.7), (0.5, 0.6),
                                         (0.4, 0.5), (0.3, 0.4), (0.2, 0.3), (0.1, 0.2), (-0.01, 0.1)])
         self.include_diagonal = True
         self.val_spectra_per_inchikey = 1
         self.random_seed: Optional[int] = None
-        self.max_pair_resampling = 1
+        self.max_pair_resampling = 10000
```
florian-huber, line 146: Why this high number?

florian-huber (approved):

ms2deepscore/SettingsMS2Deepscore.py
niekdejonge: @florian-huber Basically meant as infinite. As in there is no max. I noticed that the amount of resampling was never a lot.  And any maximum imposed, could have quite bad effect on the balance of pairs selection. I will share the analysis that came to this conclusion later today.

PR #265, opened 2025-01-29
restrict torch until new model load/save is implemented

florian-huber:


niekdejonge:
@florian-huber Thanks!

PR #294, opened 2026-06-23
Add onnx export

julianpollmann:
This adds export to onnx to the model and training script.

@florian-huber @niekdejonge does this fit or should this be added to a converter script in `ms2deepscore/models`?

Should I adapt the Pipeline to use onnxruntime for inference?


niekdejonge:
@julianpollmann Great! Thanks for developing an onnx converter and for adding it here. 

Good that you added a save option for the settings. Now it is a separate json file right. Ideally it is part of the onnx file itself, for easier use and less risk of mismatching the wrong settings with a model. I think this is possible for onnx as well, using metadata_props (suggested by claude).

For inference we would like to have matchms compatibility, which makes it a bit more work... So how it is solved now is having the class MS2DeepScore(BaseSimilarity): which inherits from the matchms BaseSimilarity to make it matchms compatible. This has some torch specific code (like eval) and takes in a SiameseSpectralModel (also torch specific). 

So what I think is needed to make it run using onnxruntime is a new MS2DeepScoreONNX class, a SiameseSpectralModelONNX class and a compute_embedding_array_onnx function, to enable inference. In fact you might be able to put all this functionality in just a single MS2DeepScoreONNX class, since SiameseSpectralModelONNX is not really needed as a separate class, since we don't do ONNX training. I think Pipeline doesn't need changing, this is on the matchms end and will just run anything that inherits from BaseSimilarity if I remember correctly, so if you give it the new SiameseSpectralModelONNX, which inherits from BaseSimilarity it should work fine with Pipeline. 

A bit more work than I realized, when I commented on your linkedin post, sorry... But I think it is a nice change worth the effort, since onnx is faster and also programming language interoperable.

julianpollmann:
@niekdejonge I will have time on friday to look into that!
Weights and settings can be integrated into onnx via metadata_props, however some providers (e.g., Huggingface) have separate config/settings.json files for e.g., model cards. Don't know which use case is more suitable for us.

Also: Right now onnx conversion is done on every checkpoint, is there a specific "last checkpoint" function or should this better be integrated into training_wrapper_functions.py?

ms2deepscore/models/SiameseSpectralModelONNX.py
```diff
@@ -0,0 +1,157 @@
...
+        -------
+        np.ndarray
+            Embedding array with shape (n_spectra, embedding_dim).
+        """
+        if batch_size <= 0:
+            raise ValueError("batch_size must be a positive integer.")
+
+        settings = self.model_settings
+
+        x_binned, x_metadata = tensorize_spectra_onnx(spectra, settings)
```
florian-huber: I would move this into the batch loop, if possible.
Just to avoid issues for super large datasets down the road.

tests/test_siamese_spectral_model_onnx.py
```diff
@@ -0,0 +1,225 @@
...
+@pytest.fixture
+def exported_onnx_model(tmp_path, model_settings):
+    """Exports a SiameseSpectralModel to ONNX and returns the path."""
+    pytorch_model = SiameseSpectralModel(model_settings)
+    pytorch_model.export_to_onnx(tmp_path, model_name="test_model")
+    return tmp_path / "test_model.onnx"
+
+
+@pytest.fixture
+def exported_onnx_model_with_metadata(tmp_path, model_settings_with_metadata):
```
florian-huber, line 76: Not sure, maybe I just overlooked this: But it would be good to also test somewhere if models without metadata (internally then handled as a tensor with width-0), also works ?

ms2deepscore/models/SiameseSpectralModel.py
```diff
@@ -160,6 +172,73 @@ def compute_embedding_array(
...
+        output_dir
+            The path to the output directory.
+        model_name
+            The name of the onnx model files (and settings).
+        """
+        out_path = Path(output_dir)
+        out_path.mkdir(parents=True, exist_ok=True)
+        self.eval()
+
+        encoder = self.encoder.cpu()
```
florian-huber, line 190: I think this could move the model in an unexpected way. Maybe better move it back to the original device after export?

florian-huber (changes requested):
Great addition, thanks @julianpollmann !
I added a few comments to check (feel free to ignore them if they are beyond the point).

tests/test_siamese_spectral_model_onnx.py
julianpollmann: In fact metadata is always exported to onnx, since it will create (static) graph. However it can have different shapes. I adjusted metadata checks in compute_embedding_array and added some tests for this.

README.md
niekdejonge, line 76: We should update this and upload the latest onnx model to a new zenodo link.

README.md
```diff
@@ -91,17 +98,43 @@ similarity_matrix = pipeline.scores.to_array()
...
 
 ## 2 Create embeddings
 
 To calculate chemical similarity scores, MS2DeepScore first calculates an embedding (vector) representing each spectrum. 
 This intermediate product can also be used to visualize spectra in "chemical space" by using a dimensionality reduction technique, like UMAP.
 
 ```python
+from ms2deepscore import MS2DeepScore, MS2DeepScoreONNX
+
 cleaned_spectra = pipeline.spectra_queries
```
niekdejonge, line 138: I think the two lines below should be removed. Since it continuous from the code above. Which is an ONNX model, so I guess this won't work with MS2DeepScore. So if we remove these two lines it should be fine.

README.md
```diff
@@ -91,17 +98,43 @@ similarity_matrix = pipeline.scores.to_array()
...
 
 ms2ds_model = MS2DeepScore(model)
 ms2ds_embeddings = ms2ds_model.get_embedding_array(cleaned_spectra)
+
+# Hardware accelerated version:
+ms2ds_model = MS2DeepScoreONNX("ms2deepscore_model.onnx")
+ms2ds_embeddings = ms2ds_model.get_embedding_array(cleaned_spectra)
 ```
```
niekdejonge, line 153: I think this will not run: 

It should be: 
ms2ds_model = MS2DeepScoreONNX(model)
or 
ms2ds_model = MS2DeepScoreONNX(SiameseSpectralModelONNX("ms2deepscore_model.onnx"))

README.md
```diff
@@ -136,6 +169,16 @@ train_ms2deepscore_wrapper(
     SettingsEmbeddingEvaluator() # this results in also training the embedding evaluator. Leave as None if you don't want to train this.
 )
 ```
+
+### Model conversion
```
niekdejonge, line 188: Add short description of why: 
In version 2.9 and earlier we used .pt models. But since ONNX models enable faster inference we changed to ONNX. You can download the latest model in onnx format from zenodo, see link at start of readme. 
If you have your own models that you would like to have converted to onnx, you can convert it using the code below.

niekdejonge (approved):
@julianpollmann Great work! Looks good to me! Very complete PR, both in tests, and readme.

The tests with the onnx model have the same output as the ones for the .pt models, so it looks all good to me!

Will be great to have the CPU speed up by ONNX, since the majority of users will just want to run it on their laptop. 

I left a few comments in the tutorial, for further clarity and some small things that won't run.

README.md
julianpollmann: Not quite sure, but I think @florian-huber is training a model on new data. Maybe we should add this one then?

README.md
julianpollmann: So you mean removing:
```
ms2ds_model = MS2DeepScore(model)
ms2ds_embeddings = ms2ds_model.get_embedding_array(cleaned_spectra)
```
?

README.md
niekdejonge: Yes that makes sense, I just want to make sure that the readme matches the actual model downloaded. Now I think the readme is for onnx but the downloaded model is pt

README.md
niekdejonge: Yes if we move fully to onnx. Otherwise I would add a bit of extra description and maybe just separate the two. Like if you have an onnx model you can run it like this. But now it is a bit mixed, which can be confusing I think.

niekdejonge: In fact it would just fail now, since we don't specify that you can even load a .pt file. The 
ms2ds_model = MS2DeepScore(model)

Would use the code in the block above: 
model = SiameseSpectralModelONNX(model_file_name)

Which I think would fail.

julianpollmann:
@niekdejonge @florian-huber I split up the Code blocks for pytorch and onnx in the README to make things clearer. For now I would keep both, since we don't have a recent model yet to convert (I could convert the latest from decembre 2025). I guess we should update the docs once we have a new model with fresh data. For now I would just merge this, if ok?

Issue #295, opened 2026-06-30
refactor/optimize embedding computation


florian-huber:
Thanks @julianpollmann for spotting this:
We so far worked with a non-batched embedding computation from single spectrum objects, which is unnecessarily slow (I think).
And, the method should probably move into the model class.

florian-huber:
Closed with #296

PR #296, opened 2026-06-30
Refactor embedding computation method(s)

florian-huber:
- move function to model-method
- improve speed by using batches

Addresses #295


julianpollmann (approved):
Looks good :)

